PR #49, opened 2016-04-19
Adding applicaicon icon for Linux

laptander:


NikolajSchlej:
Thanks a lot. Planned to do it once, but completely forgot about it afterwards.

PR #101, opened 2017-12-24
little fixies

serg-pushkarev:
1. Fix error while build action performed, becasuse TE and PE files can have empty relocation for padding. 
2. fix typo


NikolajSchlej:
Fixes integrated, teFixup issue described. Thanks for the PR.

PR #108, opened 2018-02-12
Fix uefiextract pro file

c0d3z3r0:


NikolajSchlej:
Fixed by the latest commit, so not needed anymore, but thanks for the PR.

PR #111, opened 2018-02-13
Offset base by the preceding pad file

lasalvavida:
BIOS: [Supermicro A1Sri](http://www.supermicro.com/products/motherboard/atom/x10/a1sri-2758f.cfm)

In this BIOS, pad files are inserted to align the PE32 section in PEI files. The image base in the PE header includes this, but the computed base does not currently. As a result, rebuilding any PEI module with padding will compute a negative delta (very large in unsigned) and perform a bunch of invalid relocations even though nothing has changed.

This pull request checks the preceding file, and if it is a pad file, increments the base used for rebasing by the size of the pad file.

This may be fine on its own; I don't know enough about how other BIOSes use pad files to say for sure.

### DIFF
--- a/ffsengine.cpp
+++ b/ffsengine.cpp
@@ -3921,7 +3921,15 @@ UINT8 FfsEngine::reconstructSection(const QModelIndex& index, const UINT32 base,
             }*/
 
             if (base) {
-                result = rebase(reconstructed, base - teFixup + header.size());
+                UINT16 padding = 0;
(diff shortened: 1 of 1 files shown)


NikolajSchlej:
Root cause of this bug needs to be investigated, this is just a workaround. Thanks for reporting.

matrosov:
check the latest commit it's should be fixed there

lasalvavida:
Thanks guys, confirmed fixed and closing!

PR #123, opened 2018-03-20
Set IA32_FEATURE_CONTROL.SGX_LC on Gemini Lake

jethrogb:


vit9696:
Implemented in master (https://github.com/LongSoft/UEFITool/commit/8bfa302c869fc664a9924498243ff0726cf601e8). Thanks :)

PR #142, opened 2018-08-23
Modification support & other

allowetotima:
+ Replace and rebuild functions
+ NVRAM volumes rebuild support (without changing the size of volume)
+ 'TXT' and 'Microcode' parsing tabs
+ 'Inspect with IDA' function

### DIFF
--- a/UEFITool/uefitool.cpp
+++ b/UEFITool/uefitool.cpp
@@ -760,7 +760,11 @@ void UEFITool::inspectIDA64()
 }
 
 void UEFITool::inspect(const UINT8 mode)
-{
+{ 
+    if((mode == INSPECT_MODE_IDA32 && idaPath32.trimmed().isEmpty()) ||
+            (mode == INSPECT_MODE_IDA64 && idaPath64.trimmed().isEmpty()))
+        return;
+
     UString filePath;
 
     extract(EXTRACT_MODE_BODY, &filePath);
@@ -770,10 +774,7 @@ void UEFITool::inspect(const UINT8 mode)
     QStringList arg;
     arg << filePath;
 
-    QProcess idaProcess;
-    idaProcess.setProgram(mode == INSPECT_MODE_IDA32 ? idaPath32 : idaPath64);
-    idaProcess.setArguments(arg);
-    if(!idaProcess.startDetached()) {
+    if(!QProcess::startDetached(mode == INSPECT_MODE_IDA32 ? idaPath32 : idaPath64, arg)) {
         QMessageBox::critical(this, tr("Inspect failed"), tr("Can't start IDA process"), QMessageBox::Ok);
         return;
     }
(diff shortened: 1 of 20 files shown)


vit9696:
Hi,

Um, thank you for this. The feature was slowly worked internally on, but it did not receive enough priority. With the builder being pretty much complete I guess it may be easier to just merge this and later apply the changes as needed.

I will contact the rest and review the changes when I get back home. Would you please fix CI for now? Also, on macOS and Linux ida binary is ida/ida64 without the exe suffix. Consider adding an ifdef with [Q_OS_WIN](http://doc.qt.io/qt-5/qtglobal.html#Q_OS_WIN) probably?

UEFITool/uefitool.cpp
```diff
@@ -401,7 +432,7 @@ void UEFITool::insert(const UINT8 mode)
 {
     U_UNUSED_PARAMETER(mode);
     
-    /*QModelIndex index = ui->structureTreeView->selectionModel()->currentIndex();
```
vit9696: Do you plan to add new file insertion within this pull request?

UEFITool/uefitool.cpp
```diff
@@ -858,6 +934,43 @@ void UEFITool::openImageFile(QString path)
...
+}
+
+void UEFITool::specifyPathIDA32()
+{
+    UString path;
+
+#if defined Q_OS_WIN
+    path = QFileDialog::getOpenFileName(this, tr("Specify path to ida.exe"),  "ida.exe", tr("IDA Pro executable (ida.exe);;All files (*)"));
+#else
+    path = QFileDialog::getOpenFileName(this, tr("Specify path to ida.exe"),  "ida.exe", tr("IDA Pro executable (ida);;All files (*)"));
```
vit9696: I think it is still a little messed up in the 2nd and 3rd arguments. Same below.

common/ffsbuilder.cpp
```diff
@@ -379,34 +482,966 @@ USTATUS FfsBuilder::buildFreeSpace(const UModelIndex & index, UByteArray & freeS
...
+                    if (volumeHeader->Revision > 1 && (fileHeader->Attributes & FFS_ATTRIB_LARGE_FILE))
+                        fileHeaderSize = sizeof(EFI_FFS_FILE_HEADER2);
+
+                    // Pad file
+                    if (fileHeader->Type == EFI_FV_FILETYPE_PAD) {
+                        padFileGuid = file.left(sizeof(EFI_GUID));
+
+                        // Parse non-empty pad file
+                        if (model->rowCount(index.child(i, 0))) {
+                            //TODO: handle it
```
vit9696: From what I remember, there was an issue with some firmwares expecting data in pad files. In the old version there was a warning regarding pad file contents being destroyed, might it need to be added here in addition to the todo comment, or it is already handled elsewhere?

common/ffsbuilder.cpp
```diff
@@ -379,34 +482,967 @@ USTATUS FfsBuilder::buildFreeSpace(const UModelIndex & index, UByteArray & freeS
...
+                }
+            }
+        }
+        else if(type == Types::FsysStore) {
+             UByteArray store = header + body;
+
+            // Recalculate store checksum
+            UINT32 calculatedCrc = crc32(0, (const UINT8*)store.constData(), (const UINT32)store.size() - sizeof(UINT32));
+            // Write new checksum
+            body.replace((const UINT32)body.size() - sizeof(UINT32), sizeof(UINT32), (const char*)calculatedCrc, sizeof(UINT32));
```
vit9696: I think what was meant to be written here is not `(const char*)calculatedCrc` but `&calculatedCrc`. I will fix it up in a subsequent commit, but decided to let you know in case I misunderstood something on a higher level.

PR #189, opened 2020-01-16
New engine

serg-pushkarev:


NikolajSchlej:
Any comments on why this kind of calculation is actually right, and the one we have today is wrong?

serg-pushkarev:
Source code of FitGen utility:
https://github.com/tianocore/edk2-platforms/blob/master/Silicon/Intel/Tools/FitGen/FitGen.c (line 2458).
The checksum takes into account the C_V bit (CsFlag).

PR #195, opened 2020-02-16
Enable building UEFITool as a library (QT Project Template += lib)

mtnmts:
This allows to build UEFITool as a shared library, Useful if you want to link against it.


mtnmts:
Sorry, this is broken, better probably to create another .pro file

PR #204, opened 2020-03-17
Added support for Supermicro X11 series motherboards

metacollin:
Hi, I've added patches that successfully unlock MSRs E2h and, if locked, 1AAh (this register also needs to be unlocked.  In the past, this was generally not locked, but several recent X299 and C400/C600 series motherboards tend to lock this register, and a few of the Supermicro mobos do, but not all).  

I have personally rested these patches on my own motherboard, an X11DAi-N, and several people on the insanelymac.com forums have tested them on other Supermicro BIOS images successfully.  They appear to unlock E2h on every X11 Series bios image I was able to download, several C422 Asus motherboards, and at least one Asus X999 based motherboard.


vit9696:
This looks a bit dirty to me, as the amount of bytes for patch is quite long and will likely break easily. Are you sure it cannot be done in a better way?

metacollin:
I actually intentionally made them that long because I was worried about false positives (I don't really have good intuition yet about how likely a sequence of bytes is likely to show up in a binary yet), but I think I was being way too conservative.  I pushed a change that reduces them to their core functionality as much as I dared (no false positives in the bioses I tested against at least).

The C422/X299 matches against code that does a bit test against the lock bit locations, then a bit test and set, and changes the second bit test and set to just a bit test.  The function is identical across multiple BIOSes, but I simplified it further.  The jump address is the match any (....) bytes of the patch.  

Example, with the second bts already changed to a b (0x0D of course is the lock bit for the 0x1AA MSR):  

Patched ASM

The C620 patches simply zeros out the hard coded lock value (0x8000 or 0x2000 depending on which MSR is being locked)  as they literally just or the MSR with that hard coded value:

Match ASM ==> Patch ASM

vit9696:
48h offset on the stack is a bit vague, but I guess it will unlikely change for this platform in the future. Merged.

CrazyHulk:
Did this patch works on ASROCK epc621d8a which is C621.

@metacollin

PR #211, opened 2020-07-11
unixbuild.sh: sometimes bash lives elsewhere

nms42:


nms42:
Drop it. One can always do

bash unixbuild.sh

PR #224, opened 2020-11-02
parsing new Bootguard structures on 10 gen CPU

prop263:
Hi Alex
New BG ACM has a lot of redesign and the KEY and IBB manifests were changed.
I reversed some of the stuff and added initial support so the 'security tab' shows good info.
Sorry for the ugly duplicate code, best way was to overload it with virtual function but i wanted to make minimal changes.
Also i realized that the new design supports for RSA 3072 bit so in the next stage we will need also update the RSA structs. Till now i haven't seen a KeyM with such keys.
Danny

### DIFF
--- a/common/bootguard.h
+++ b/common/bootguard.h
@@ -121,11 +121,29 @@ typedef struct BG_BOOT_POLICY_MANIFEST_HEADER_ {
     UINT16 NEMDataSize;
 } BG_BOOT_POLICY_MANIFEST_HEADER;
 
+typedef struct BG_BOOT_POLICY_MANIFEST_HEADER2_ {
+    UINT64 Tag;
+    UINT8  Version;
+    UINT8  HeaderVersion;
+    UINT16 HeaderSize;
+    UINT16 RSAEntryOffset;
+    UINT8  PMBPMVersion;
+    UINT8  BPSVN;
+    UINT8  ACMSVN;
+    UINT8  : 8;
+    UINT16 NEMDataSize;
+} BG_BOOT_POLICY_MANIFEST_HEADER2;
+
 typedef struct SHA256_HASH_ {
     UINT16 HashAlgorithmId;
     UINT16 Size;
     UINT8  HashBuffer[32];
 } SHA256_HASH;
+
+typedef struct HASH_HEADER_ {
+    UINT16 HashAlgorithmId;
+    UINT16 Size;
+} HASH_HEADER;
 
 typedef struct RSA_PUBLIC_KEY_ {
     UINT8  Version;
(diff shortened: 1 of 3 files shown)


common/bootguard.h
```diff
@@ -180,6 +198,30 @@ typedef struct BG_IBB_ELEMENT_ {
     // BG_IBB_SEGMENT_ELEMENT IbbSegment[];
 } BG_IBB_ELEMENT;
 
+typedef struct BG_IBB_ELEMENT2_ {
+    UINT64                 Tag;
+    UINT8                  Version;
+    UINT8                  : 8;
```
vit9696, line 204: This looks ugly to me, why cannot you put Unknown or something alike and avoid the bit count, which is the default for UINT8 anyway?

common/ffsparser.cpp
```diff
@@ -3778,12 +3779,23 @@ USTATUS FfsParser::parseFit(const UModelIndex & index)
                     break;
 
                 case FIT_TYPE_AC_KEY_MANIFEST:
-                    status = parseFitEntryBootGuardKeyManifest(item, localOffset, itemIndex, info, currentEntrySize);
+                    manifestType = (const char*)(item.constData() + localOffset);
+                    //check type
+                    if (manifestType[0x8]>=0x20)
```
vit9696, line 3784: This needs bounds checking really...

common/ffsparser.cpp
```diff
@@ -3778,12 +3779,23 @@ USTATUS FfsParser::parseFit(const UModelIndex & index)
...
+                    else
+                        status = parseFitEntryBootGuardKeyManifest(item, localOffset, itemIndex, info, currentEntrySize);
                     kmIndex = itemIndex;
                     break;
 
                 case FIT_TYPE_AC_BOOT_POLICY:
-                    status = parseFitEntryBootGuardBootPolicy(item, localOffset, itemIndex, info, currentEntrySize);
+                    manifestType = (const char*)(item.constData() + localOffset);
+                    //check type
+                    if (manifestType[0x8]>=0x20)
```
vit9696, line 3794: Same here.

common/ffsparser.cpp
```diff
@@ -4088,6 +4168,215 @@ USTATUS FfsParser::findNextBootGuardBootPolicyElement(const UByteArray & bootPol
...
+
+                currSHA = currSHA + sizeof(HASH_HEADER) + currSHASize;
+            }
+
+
+            // Add all IBB segments
+            securityInfo += UString("\n\nIBB Segments:\n");
+            const BG_IBB_SEGMENT_ELEMENT* segments = (const BG_IBB_SEGMENT_ELEMENT*)((char*)elementHeader + sizeof(BG_IBB_ELEMENT2) + elementHeader->SizeOfDigests +3);
+            UINT8 IbbSegCount = *((char*)segments -1);
+            for (UINT8 i = 0; i < IbbSegCount; i++) {
```
vit9696, line 4280: This would also want bounds checking.

vit9696:
There seem to be several issues with bounds checking which need to be addressed before merging. I marked some, and the rest are very similar. There is Windows CI failing, but it seems to be caused by msys2 issues. It is best to fix it prior to merging as well.

PR #226, opened 2020-11-09
Added security info to UEFIExtract

krsh:


vit9696:
This looks about right to me, but I a, worried whether it is going to break any automated parsers.

NikolajSchlej:
Those parsers will need to be fixed then. Approved.

PR #230, opened 2020-11-27
Added ability to dump files found with search

s3rb31:
This might be a useful addition. I was really annoyed I couldn't simply dump all the files which are returned with a text search so I implemented it. Pretty ad-hoc implementation but it does what it's supposed to.

I only tried this on GNU/Linux but I see no reason why it shouldn't be working on Windows. At the moment I don't have access to a Windows machine so I cannot test it but I will build and test this on Windows once I get home (next week probably).

If you think this functionality improves UEFITool but you want changes to the code feel free to tell me and I'll adjust the PR accordingly.

Greetings

PS. There is an explanation with screenshots availabe at my fork. Have a look [here](https://github.com/s3rb31/UEFITool/blob/new_engine-mod/README.md).

### DIFF
--- a/UEFITool/ffsfinder.cpp
+++ b/UEFITool/ffsfinder.cpp
@@ -57,6 +57,11 @@ USTATUS FfsFinder::findHexPattern(const UModelIndex & index, const UByteArray & 
             // For patterns that cross header|body boundary, skip patterns entirely located in body, since
             // children search above has already found them.
             if (!(hasChildren && mode == SEARCH_MODE_ALL && offset/2 >= model->header(index).size())) {
+                UModelIndex savedIndex = index;
+                if (model->type(savedIndex) == Types::Section)
+                    savedIndex = index.parent();
+                if (model->type(savedIndex) == Types::File)
+                    addFoundFile(savedIndex);
                 msg(UString("Hex pattern \"") + UString(hexPattern)
                     + UString("\" found as \"") + hexBody.mid(offset, hexPattern.length()).toUpper()
                     + UString("\" in ") + model->name(model->parent(index))
@@ -126,6 +131,11 @@ USTATUS FfsFinder::findGuidPattern(const UModelIndex & index, const UByteArray &
     INT32 offset = regexp.indexIn(hexBody);
     while (offset >= 0) {
         if (offset % 2 == 0) {
+            UModelIndex savedIndex = index;
+            if (model->type(savedIndex) == Types::Section)
+                savedIndex = index.parent();
+            if (model->type(savedIndex) == Types::File)
+                addFoundFile(savedIndex);
             msg(UString("GUID pattern \"") + UString(guidPattern)
                 + UString("\" found as \"") + hexBody.mid(offset, hexPattern.length()).toUpper()
                 + UString("\" in ") + model->name(model->parent(index))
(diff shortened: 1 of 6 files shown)


vit9696:
There is a CI failure that needs to be addressed first of all. I wonder if a checkbox is an adequate interface for this action. Cannot think of anything better, but we have UEFIFind tool that does exactly that…

s3rb31:
> There is a CI failure that needs to be addressed first of all. 

Actually the code compiles with modern g++, clang++ (with c++0x enabled) and MSVC. The question is why aren't c++0x features enabled for the linux CI task but for Mac and Windows? This builds fine with g++ 10.2.0 on archlinux without any modifications to the `unixbuild.sh` script. 

I cannot do anything about the Coverity CI task since it is failing to fetch some script with error `401`.

> I wonder if a checkbox is an adequate interface for this action. Cannot think of anything better, 

I could imagine a menu-entry for this that is enabled after a successful search. This way it would also be easier to implement this functionality for all search types (Hex, GUID, Strings). I like this idea.

> but we have UEFIFind tool that does exactly that…

No. UEFIFind has no concept of strings an thus does not support case insensitive search. To search for a string I'd have to generate `2^n` different patterns to cover all possible variations of upper/lowercase representations of a string. Not really practical.

Also UEFIFind cannot directly extract the files that contain occurrences of the pattern, so you have to take it's output, filter for duplicates and then create some script that funnels the output into `UEFIExtract`. I tried it and in the end just opted to implement this feature into UEFITool, much less headache. Also UEFIExtract is not using the GUID database to convert the GUIDs of the dumped files to useful names.

It might be worth to think about implementing string-search (with the ability to do case-insensitive searches) and extraction functionality into UEFIFind if you don't like this solution. For now I'll just move the checkbox to a menu-entry and enable this functionality for all search types.

Regarding the use of C++0x features: You want to stick with this old g++ version used in CI? If yes I'll change the code but I really believe it would be better to just update the g++ version CI is using.

vit9696:
@NikolajSchlej had some strong concerns for not using newer C++ versions due to some legacy compiler support in the past, but I guess at this point it should no longer be relevant. Perhaps wait for him to comment on this case. As for Coverity, their servers are indeed down at the moment.

> I could imagine a menu-entry for this that is enabled after a successful search. This way it would also be easier to implement this functionality for all search types (Hex, GUID, Strings). I like this idea.

That is actually an interesting idea. Context menu in the results, right?

s3rb31:
> > I could imagine a menu-entry for this that is enabled after a successful search. This way it would also be easier to implement this functionality for all search types (Hex, GUID, Strings). I like this idea.
> 
> That is actually an interesting idea. Context menu in the results, right?

I was thinking about something like this:

example

s3rb31:
Okay, so I changed the range-based `for`-loop to Qt's `foreach` and adjusted the enum references so it passes the CI checks now.

I enabled support for all three search types and changed the interface to look like this:

UEFITool-mod-1

The menu entry is only enabled after a successful search and stays enabled until the next failed search. Only the files from the last search will be extracted.

I hope you like this.

NikolajSchlej:
I still don't see why it needs to be in UEFITool, because for mass extraction we have UEFIExtract and for searching in binary files we have various kinds of grep. Extracting everything and then searching only in the leaves of the extracted tree does the same without creating new UI for a fairly niche operation (that is somehow about files only).

NikolajSchlej:
If we do want the "Actions for search results", which would be something like "Copy, Clear, Extract all as is..., Extract all bodies..., etc.", that should likely be separated into a new "Search" menu, with the current "Action->Search..." moved there. Also, if we want to search in whatever types/subtypes, that should be an option in "Search" window.

I would like this PR to be a source of inspiration for that Search improvement, but do not yet want to merge it in the current state. Still really appreciate it.

NikolajSchlej:
Closing this to create an issue instead.

PR #237, opened 2021-04-03
UEFITool NE 59

vampirecat35:
- Updates for QT 5 and QT 6
- Deprecation fixes
- Code fixes


vit9696:
Hello, please update the compatibility with the CI. We still require Qt 5.6 as a minimum for LGPL 2.0 license reasons.

vampirecat35:
> Hello, please update the compatibility with the CI. We still require Qt 5.6 as a minimum for LGPL 2.0 license reasons.

I used Qt-5.15.2 and Qt-6.0.3
I tested it on macOS.

vampirecat35:
> Hello, please update the compatibility with the CI. We still require Qt 5.6 as a minimum for LGPL 2.0 license reasons.

Done.
I added a QT version check.
If 5.11 or older use .width
Otherwise use new method.

.github/workflows/main.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+x@name: CI
```
vit9696, line 1: You broke CI completely here by accident.

UEFITool/uefitool.pro
```diff
@@ -8,6 +8,8 @@ DEFINES += "U_ENABLE_NVRAM_PARSING_SUPPORT"
 DEFINES += "U_ENABLE_ME_PARSING_SUPPORT"
 DEFINES += "U_ENABLE_GUID_DATABASE_SUPPORT"
 
+greaterThan(QT_MAJOR_VERSION, 5): DEFINES += "QT_VERSION_MAJOR=6"
```
vit9696, line 11: Are you sure there is no builtin macro to check against Qt version?

.github/workflows/main.yml
vampirecat35: Fixed.
Accidentally edited the file.

vampirecat35:
CI is fixed and build success.
Also tested on newer QT versions and success too.

UEFITool/uefitool.pro
vampirecat35: True.
This isn't needed.
Pushed commit

UEFITool/ffsfinder.cpp
```diff
@@ -29,7 +29,12 @@ USTATUS FfsFinder::findHexPattern(const UModelIndex & index, const UByteArray &
 
     bool hasChildren = (model->rowCount(index) > 0);
     for (int i = 0; i < model->rowCount(index); i++) {
+#ifdef _USE_DEPRECATED
```
vit9696, line 32: Why cannot we use non-deprecated interfaces exclusively? Does Qt 5.6 not support them? Who does define `_USE_DEPRECATED`?

vampirecat35: I'll do a version check.
Fixing in a minute

vampirecat35:
Fixed older QT versions.
Newer ones will use the new code.

common/treemodel.h
```diff
@@ -86,7 +88,7 @@ class UModelIndex
 #if defined(QT_CORE_LIB)
 class TreeModel : public QAbstractItemModel
 {
-    Q_OBJECT
+    //Q_OBJECT
```
vit9696, line 91: Better just remove than comment out.

common/treemodel.h
```diff
@@ -35,7 +35,9 @@ enum ItemFixedState {
 #include "types.h"
 #include "treeitem.h"
 
+#ifndef UModelIndex
```
vit9696, line 38: Why is that one needed?

common/ustring.cpp
```diff
@@ -19,7 +19,7 @@ UString usprintf(const char* fmt, ...)
     UString msg;
     va_list vl;
     va_start(vl, fmt);
-    msg.vsprintf(fmt, vl);
+    msg = msg.vasprintf(fmt, vl);
```
vit9696, line 22: Why is this necessary?

vit9696:
Looks mostly good, spotted a few more weird points, but should be ready to merge soon after.

vampirecat35:
I added it using a version check.
This way older than QT 5.6 also builds.

common/ustring.cpp
vampirecat35: Newer QT versions like QT 6 don't implement .vsprintf.
This thus needs the newer method.

common/treemodel.h
vampirecat35: Removing...

common/treemodel.h
vampirecat35: Removed

vampirecat35:
Image of UEFITool on macOS Big Sur using QT 6
[https://i.ibb.co/WnFR9YZ/UEFITool-NE-59-mac-OS-QT6.png](UEFITool QT6)

vampirecat35:
Removed Xcode projects, copied new icon file, added uefitool.entitlements for use with code sign
codesign -fs "Apple Development" --entitlements uefitool.entitlements was used in my case.
This is needed after macdeployqt for adding sandbox and signing.

vampirecat35:
Done with all QT 6 updates.
It now builds without a warning for QT 6.

PR #241, opened 2021-07-19
New engine fixes

joevt:
While looking at firmware `ROG-STRIX-Z590-E-GAMING-WIFI-ASUS-0902.CAP`, I found some issues with UEFIExtract. I found more issues by comparing output from UEFIExtract and UEFITool.


joevt:
Hopefully the last commit will be compilable in Visual Studio.

joevt:
We might want to remove the hash size messages or maybe only include a message if the size is not the old 0x20 or the new 0x30 length? Or don't output a message at all as long as the hash size matches the size specified in the structure? Is there a spec that shows what the possible valid lengths are?

UEFIExtract/ffsdumper.cpp
```diff
@@ -15,22 +15,52 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 
 #include <fstream>
 
+static void FixFileName(UString &name)
```
NikolajSchlej, line 18: Please make it into FfsDumper::fixFileName.

UEFIExtract/ffsdumper.cpp
```diff
@@ -15,22 +15,52 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
...
+
 USTATUS FfsDumper::dump(const UModelIndex & root, const UString & path, const DumpMode dumpMode, const UINT8 sectionType, const UString & guid)
 {
     dumped = false;
     counterHeader = counterBody = counterRaw = counterInfo = 0;
     fileList.clear();
 
-    if (changeDirectory(path))
+    if (changeDirectory(path)) {
+        fprintf(stdout, "Can't change directory %s\n", (const char*)path.toLocal8Bit());
```
NikolajSchlej, line 51: Case and presence/absence of ":" is inconsistent in this new error messages. 
Also, why not just printf(...) instead of fprintf(strdout, ...) ?

printf("Something is wrong: %u", error_code) should be the consistent format, IMO.

UEFITool/uefitool.pro
```diff
@@ -123,7 +123,5 @@ FORMS += uefitool.ui \
 RESOURCES += uefitool.qrc
 RC_FILE = uefitool.rc
 ICON = icons/uefitool.icns
-ICONFILE.files = icons/uefitool.icns
-ICONFILE.path = Contents/Resources
```
NikolajSchlej, line 127: AFAIR, this was for Qt4, but we don't really support it, so I'm OK with this change.

NikolajSchlej (changes requested):
Made some nitpicks, but otherwise LGTM.

UEFIExtract/ffsdumper.cpp
joevt: fixed in new commit.

UEFIExtract/ffsdumper.cpp
joevt: fixed in new commit https://github.com/LongSoft/UEFITool/pull/241/commits/4c35a9d670ae0478424d5cc2e42ada424a0722e4

joevt:
These commits should address the comments and change request.
There's a couple new things I found while trying to compile for latest versions of QT5 and QT6. I hope they'll work for other build environments (mine is macOS).

UEFIExtract/ffsdumper.h
```diff
@@ -45,6 +45,7 @@ class FfsDumper
 
 private:
     USTATUS recursiveDump(const UModelIndex & root, const UString & path, const DumpMode dumpMode, const UINT8 sectionType, const UString & guid);
+	void fixFileName(UString &name);
```
NikolajSchlej, line 48: Something strange with whitespace

common/ffsparser.cpp
```diff
@@ -164,7 +160,7 @@ USTATUS FfsParser::parseGenericImage(const UByteArray & buffer, const UINT32 loc
 {
     // Parse as generic UEFI image
     UString name("UEFI image");
-    UString info = usprintf("Full size: %Xh (%u)", buffer.size(), buffer.size());
+    UString info = usprintf("Full size: %" PRIXQ "h (%" PRIuQ ")", buffer.size(), buffer.size());
```
NikolajSchlej, line 163: I know this things are better, but they are also ugly as all hell.

common/qinttypes.h
```diff
@@ -0,0 +1,38 @@
+/* qinttypes.h
+
+Copyright (c) 2021, Nikolaj Schlej. All rights reserved.
```
NikolajSchlej, line 3: Lol

NikolajSchlej (approved):
Some more nitpicks, but fine otherwise. Probably need to call that qinttypes.h uinttypes.h or something, and include the Qt one inside it in Qt builds.

common/ffsparser.cpp
joevt: Alternative is to typecast to UINT32 as was done in 34c8ad8dcc5f4bd76509123626d0f00511d3eb45.
Or use a different function for printing output, such as the {fmt} library https://github.com/fmtlib/fmt

joevt:
I'll rename qintypes.h to uinttypes.h to match ubytearray.h and ustring.h

There's no Qt version of the file. inttypes.h is a standard C/C++ include. uinttypes.h exists just to define format macros similar to the ones defined in inttypes.h which will work for Qt builds and non-Qt builds and for different versions of Qt (5 and 6 for example). Maybe the #if statement in uinttypes.h needs to be modified for different environments that I have not tested.

Before these commits, UEFITool was already using the PRIX64 format macro in ffsparser.cpp and nvramparser.cpp (and still does).

NikolajSchlej:
> Before these commits, UEFITool was already using the PRIX64 format macro in ffsparser.cpp and nvramparser.cpp (and still does).

Not arguing with that, it's still ugly as all hell: "a:%"PRIX64"h" two useless punctuation symbols, ALL CAPS BECAUSE MACROS ARE FUN. "a:%Xh" is way more readable to me. Yet this "new" way is better for cross-compiler and cross-architecture compatibility, so I'm ultimately fine with it.

UEFIExtract/ffsdumper.cpp
```diff
@@ -15,22 +15,52 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
...
+void FfsDumper::fixFileName(UString &name)
+{
+    // Replace some symbols with underscores for compatibility
+    const char table[] = {
+        '/', // Banned in *nix and Windows
+        '<', '>', ':', '\"', '\\', '|', '?', '*', // Banned in Windows
+    };
+    int nameLength = (int)name.length(); // Note: Qt uses int for whatever reason.
+    if (!nameLength) {
+        name += '_';
```
vit9696, line 27: `name = "_";` is more readable as it shows the actual intention of the code — to always provide some filename.

UEFITool/uefitool.cpp
```diff
@@ -647,6 +647,38 @@ void UEFITool::openImageFile(QString path)
...
+    }
+
+    // Get last VTF
+    std::vector<std::pair<std::vector<UString>, UModelIndex > > fitTable = ffsParser->getFitTable();
+    if (fitTable.size()) {
+        printf("%s\n", "---------------------------------------------------------------------------");
+        printf("%s\n", "     Address      |   Size    |  Ver  | CS  |          Type / Info          ");
+        printf("%s\n", "---------------------------------------------------------------------------");
+        for (size_t i = 0; i < fitTable.size(); i++) {
+            printf("%s%s%s%s%s%s%s%s%s%s%s\n", (const char*)fitTable[i].first[0].toLocal8Bit(), " | "
```
vit9696, line 664: Feels really weird to me to add `, " | "` through an argument. Perhaps make it part of the formatted string?

UEFITool/uefitool.pro
```diff
@@ -9,6 +9,7 @@ DEFINES += "U_ENABLE_ME_PARSING_SUPPORT"
 DEFINES += "U_ENABLE_GUID_DATABASE_SUPPORT"
 
 HEADERS += uefitool.h \
+ ../common/uinttypes.h \
```
vit9696, line 12: Move common file to a common group please. E.g. near `../common/ubytearray.h`.

common/bstrlib/bstrlib.h
```diff
@@ -29,6 +29,12 @@ extern "C" {
 # endif
 #endif
 
+#if defined(_MSC_VER)
+#define _ATTRIBUTE_FORMAT_(t,f,a)
```
vit9696, line 33: Formatted string is a GNU extension, that is supported by selected compilers, such as gcc or clang. When clang works in MSVC-compatible mode it will set `_MSC_VER` but will still provide the extension.

In general, I believe this code is still incorrect as it makes an exception for a known to be unsupported compiler, but does not make exceptions for unknown compilers that also are unsupported. You should not start the identifiers with underscore as well. Please change to:

```C
#if defined(__clang__) || defined(__GNUC__)
#define ATTRIBUTE_FORMAT_(t,f,a) __attribute__((format(t, f, a)))
#else
#define ATTRIBUTE_FORMAT_(t,f,a)
#endif
```

common/bstrlib/bstrwrap.h
```diff
@@ -75,6 +75,12 @@
 
 #endif
 
+#if defined(_MSC_VER)
```
vit9696, line 78: Do not make duplicates here. `bstrwrap.h` includes `bstrlib.h` immediately, there is no need for this macro addition here.

common/ffsparser.cpp
vit9696: I generally recommend type casts, because we have predictable sizes of the values. Perhaps not to `UINT32` but to `unsigned`, because `UINT32` can be `unsigned long` or `unsigned int` depending on the compiler (e.g. gcc vs clang on i386).

common/ffsparser.cpp
```diff
@@ -4937,13 +4928,27 @@ USTATUS FfsParser::parseCpdRegion(const UByteArray & region, const UINT32 localO
...
+                    msg(usprintf("%s: CPD partition can't fit into its region, truncated", __FUNCTION__), partitions[i].index);
+                }
                 partitions[i].ptEntry.Length = (UINT32)region.size() - (UINT32)partitions[i].ptEntry.Offset.Offset;
             }
         }
 
         // Check for intersection with previous partition
         if (partitions[i].ptEntry.Offset.Offset < previousPartitionEnd) {
+            // Check if previous partition was compressed but did not have metadata
+            if (!partitions[i - 1].hasMetaData && partitions[i - 1].ptEntry.Offset.HuffmanCompressed) {
```
vit9696, line 4945: Is it safe to subtract 1 here?

common/ustring.h
```diff
@@ -25,7 +25,13 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 #define UString CBString
 #endif // QT_CORE_LIB
 
-UString usprintf(const char* fmt, ...);
+#if defined(_MSC_VER)
```
vit9696, line 28: Cannot we reuse that macro from bstrlib? If we cannot, I believe it should be put to a separate header.

common/utility.cpp
```diff
@@ -107,6 +107,7 @@ UString uniqueItemName(const UModelIndex & index)
         for (size_t j = 0; j < sizeof(table); j++) {
             if (name[i] == table[j]) {
                 name[i] = '_';
+                break;
```
vit9696, line 110: This code is very similar to fixFileName above. Can we reuse it?

vit9696 (commented):
Thanks for the improvements! Left some notes inline.

NikolajSchlej:
@vit9696, if UINT32 is not 32 bit, it's a problem by itself, but %X and %u should cover both unsigned int and unsigned long, because they are guaranteed to support at least 32 bits of values, so casting to UINT32 is perfectly justified, IMO. 
I also agree on the fact that default formatted output in C and C++ is shitty anyway, and a better solution like fmt library should likely be used. Back before making non-Qt versions of commndline tools I've uses Qt-specific tr().arg() and it was way easier to handle.

vit9696:
@NikolajSchlej the standard disagrees with this, and there actually is practical reasoning behind this. %lu and %u stand for unsigned long and unsigned int correspondingly. As a result there are two cases which explain why neither can be used to print UINT32 without a cast:
- According to the standard unsigned int is only guaranteed to at least be ~16-bit. Therefore it is possible for %u to be less than UINT32 and thus the behaviour is undefined as we may be passing larger integers to smaller formal arguments without a cast.
- On many 64-bit targets unsigned long is 64-bit. As a result, whether or not passing a 32-bit integer to %lu will be interpreted correctly is undefined. In practice it depends on the ABI — on System V passing 7th 32-bit integer will result in stack use for argument parsing, which means we may read 4 more bytes or rubbish.

It is fairly safe to use %lu and cast to unsigned long on most targets.

NikolajSchlej:
There are zero platforms we support or will support that still have 16 bit unsigned int, so standard can safely go to the same hell where that can still be true.

UEFIExtract/ffsdumper.h
joevt: fixed in commit f71da12ec19f86d842648eeef3e2b4580c4f99d5

UEFITool/uefitool.cpp
joevt: I wanted UEFITool to output to stdout what UEFIExtract does so that their execution/outputs can be compared more easily. But UEFIExtract uses iostream (cout::stdout) while UEFITool just uses printf. I forget if I tried iostream in UEFITool. Anyway, the printf calls look like that so that they look like the cout calls used by uefiextract_main.cp .

Is it ok to use iostream in UEFITool? If so, then I could do that and maybe make a common function that both UEFITool and UEFIExtract can use to output this info.

If we can't use iostream in UEFITool, then an alternative would be to replace cout with printf in UEFIExtract. That way I can still make both use a common function.

common/ffsparser.cpp
joevt: The index variable `i` starts at 1 in the for loop so there is always a `partitions[i - 1]` to compare with `partition[i]`.

UEFITool/uefitool.cpp
vit9696: I do not mind iostream in both, I think we would agree with @NikolajSchlej that both interfaces (iostream and printf) are not great in C++.

common/ffsparser.cpp
joevt: I think I prefer the format macros. I understand all sizes will probably be less than 32 bit but I don't like typecasting to a smaller size if there's an alternative.

If you decided to go with type cast, I would probably use the separate header uinttypes.h to define the type that you want to type cast to. Something like:
`typedef unsigned QINT`

I think `unsigned` is the correct type to use because the lack of a size qualifier is similar to the lack of size qualifier in the printf formats `%X` and `%u`.

If you want to make sure `unsigned` is always greater than 16 bits, then you can add a statement that would cause a compiler error.
`char testuinttypes[sizeof(unsigned) >= 4 ? 1 : -1];`

UEFIExtract/ffsdumper.cpp
joevt: fixed in 2926a9ff4625d10989bc2713517a0fe9f395abdf

UEFITool/uefitool.pro
joevt: fixed in 582910fb96edefffa7c51442cc39681f90ece010

common/bstrlib/bstrlib.h
joevt: Fixed in 582910fb96edefffa7c51442cc39681f90ece010

common/bstrlib/bstrwrap.h
joevt: Fixed in 582910fb96edefffa7c51442cc39681f90ece010

common/utility.cpp
joevt: Fixed in 2926a9ff4625d10989bc2713517a0fe9f395abdf

vit9696:
Merged.

PR #242, opened 2021-09-09
Add some more GUID definitions to guids.csv.

assafcarlsbad:
Courtesy of https://github.com/theorbtwo/ida-namer/blob/master/ida-namer.py


NikolajSchlej:
@vit9696, this one needs a followup to remove prefixes like g (this is just a TianoCore way to call global variables, not a part of an actual name), file_* GUID_* and so on. That's why I kind of tried to handle this file manually instead of dumping the output of some GUID collection tools and be done with it.

vit9696:
@NikolajSchlej, yes, noticed & fixed. Thx

PR #244, opened 2021-09-11
Add even more GUIDs!

assafcarlsbad:
Courtesy of https://github.com/3vilWind/diopside/blob/master/data/guids-db.ini


vit9696:
I think these GUIDs will not happen to be used as file identifiers, since these are mostly protocol GUIDs.

Issue #216, opened 2020-07-29
Add support for FPT Header version 0x21 with crc32 checksum (Intel TigerLake+)


nms42:
No patch yet

nms42:
Author of MEAnalyzer wrote to me: version 0x21 will be in firmware for TigerLake and following.

NikolajSchlej:
Need a BIOS image sample here before we go change the code.

nms42:
There are no BIOSes with 0x21 headers yet.

NikolajSchlej:
Let's wait for such a sample then before changing the code.

ValZapod:
There are samples in Intel leak. Crazy amount of them.

NikolajSchlej:
Found an image with FPT 0x21 header, uploading for future reference.
[ROG-MAXIMUS-Z690-HERO-ASUS-1720.CAP.zip](https://github.com/LongSoft/UEFITool/files/9421773/ROG-MAXIMUS-Z690-HERO-ASUS-1720.CAP.zip)

NikolajSchlej:
An image with FPT 0x20 header to test the other code branch.
[MAXIMUS-IV-GENE-Z-GEN3-ASUS-3603.ROM.zip](https://github.com/LongSoft/UEFITool/files/9422214/MAXIMUS-IV-GENE-Z-GEN3-ASUS-3603.ROM.zip)

NikolajSchlej:
Resolved in A60, will be released soon.

PR #274, opened 2022-08-25
Add support for Intel ME FPT header version 2.1

NikolajSchlej:
Also adds some missing JEDEC IDs for XMC flash chips (a fully-owned Chinese branch of Micron)
XMC

Also fixes some minor warnings from ME parser by adding missing BPDT entry types


NikolajSchlej:
Resolves #216.

Issue #278, opened 2022-08-28
Fails to compile with QT6


hughsie:
QT5 works fine, QT6 fails with:

	../common/nvramparser.cpp: In member function ‘USTATUS NvramParser::parseNvarStore(const QModelIndex&)’:
	../common/nvramparser.cpp:95:51: error: expected ‘)’ before ‘__PRI_64_LENGTH_MODIFIER__’
	   95 |             UString info = usprintf("Full size: %" PRIXQ "h (%" PRIuQ ")", padding.size(), padding.size());

NikolajSchlej:
Compiler type and version, exact Linux variant, anything to ease reproduction of it?

NikolajSchlej:
Reproduced, fixing.

NikolajSchlej:
Fixed.

PR #279, opened 2022-08-28
Make format strings sane again

NikolajSchlej:
Also re-indent all CPP files that aren't libraries to make indentation consistent across the whole thing.


NikolajSchlej:
Also fixes #278, local testing on Ubuntu arm64 with Qt6.3 and on macOS 13 beta with Qt6.4 produces no warnings for the files in this repo.

PR #284, opened 2022-08-31
Add Meson buildsystem

hughsie:
This allows UEFIExtract to build on a greater variety of targets and more
importantly allows us to build with the system-defined hardening protections
present in enterprise distributions.

### DIFF
--- a/UEFIExtract/meson.build
+++ b/UEFIExtract/meson.build
@@ -1,0 +1,18 @@
+executable(
+  'UEFIExtract',
+  sources: [
+    'uefiextract_main.cpp',
+    'ffsdumper.cpp',
+    'uefidump.cpp',
+  ],
+  link_with: [
+    lzma,
+    bstrlib,
+    uefitoolcommon,
+  ],
+  dependencies: [
+    zlib,
+  ],
+  install: true,
+  install_dir: get_option('bindir')
+)
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,11 @@
+project('UEFITool', ['c', 'cpp'],
+  version: 'A60',
+  license: 'BSD-2-Clause',
+  meson_version: '>=0.53.2',
+  default_options : ['c_std=c11', 'cpp_std=c++11'],
+)
+
+zlib = dependency('zlib')
+
+subdir('common')
+subdir('UEFIExtract')
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,6 +75,18 @@ jobs:
           tag: ${{ github.ref }}
           file_glob: true
 
+  build_linux_meson:
+    name: Build on Linux with Meson
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get Deps
+        run: sudo apt-get install -qq zlib1g-dev meson qt5-default
+      - name: Configure build
+        run: mkdir build-meson && meson ./build-meson
+      - name: Build everything
+        run: ninja -C build-meson
+
   build_win:
     name: Build on Windows
     runs-on: windows-2019
--- a/common/meson.build
+++ b/common/meson.build
@@ -1,0 +1,41 @@
+lzma = static_library('lzma',
+  sources: [
+    'LZMA/LzmaDecompress.c',
+    'LZMA/SDK/C/Bra86.c',
+    'LZMA/SDK/C/LzmaDec.c',
+    'Tiano/EfiTianoDecompress.c',
+  ],
+)
+
+bstrlib = static_library('bstrlib',
+  sources: [
+    'bstrlib/bstrlib.c',
+    'bstrlib/bstrwrap.cpp',
+  ],
+)
+
+uefitoolcommon = static_library('uefitoolcommon',
+  sources: [
+    'guiddatabase.cpp',
+    'types.cpp',
+    'descriptor.cpp',
+    'ffs.cpp',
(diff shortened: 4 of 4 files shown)


vit9696:
- Perhaps this should build at least other tools?
- Can this be added to the CI?

hughsie:
> Perhaps this should build at least other tools?

It certainly /could/ -- but I wanted to test the waters first.

> Can this be added to the CI?

Also it could, but it depends if another buildsystem is terribly interesting to the maintainers. :)

NikolajSchlej:
Well, if we are to add something into the mainline repo, it becomes interesting to the maintainers one way or another, because it's for them to maintain then. I have no issue with having different kinds of makefiles, other than having to support them all at once and make sure than any changes (i.e. new files for new BootGuard parser I'm currently working on) are made in all of them. This is why adding Meson to CI is a requirement @vit9696 mentioned, because otherwise Meson builds will be undertested, so eventually we will break them without noticing.

NikolajSchlej:
As for building every target in new_engine repo, we can start with UEFIExtract as the most useful tool for various kinds of automation, and see if anybody expresses any need to build UEFITool (qmake is fine for it because of Qt being mandatory anyway) or UEFIFind (CMake, mostly used by one project - UBU) using Meson.

meson.build
```diff
@@ -0,0 +1,11 @@
+project('UEFITool', ['c', 'cpp'],
+  version: 'A60',
+  license: 'BSD-2-Clause',
+  meson_version: '>=0.60.0',
```
vit9696: Is there any good reason to require such a recent version of meson?

hughsie: None whatsoever. I'll put it down to 0.53.2.

NikolajSchlej:
LGTM, merging.

meson.build
eli-schwartz: As far as I can tell, you are not currently requiring any features that require specific versions of Meson anyway?

UEFIExtract/meson.build
```diff
@@ -0,0 +1,18 @@
...
+  link_with: [
+    lzma,
+    bstrlib,
+    uefitoolcommon,
+  ],
+  dependencies: [
+    zlib,
+  ],
+  install: true,
+  install_dir: get_option('bindir')
```
eli-schwartz, line 17: bindir is already the default for executables. :)

meson.build
vit9696: Can we support at least 0.37 to allow for ubuntu 18.04 and debian 9?

eli-schwartz: Probably. That's the oldest version for which Meson emits feature compatibility warnings when using functionality too new / old and deprecated, and this project issues no warnings when pinning to that minimum.

eli-schwartz: Actually decided to test this. 0.37.0 fails to build:

```
ccache c++  -o UEFIExtract/UEFIExtract 'UEFIExtract/UEFIExtract@exe/uefiextract_main.cpp.o' 'UEFIExtract/UEFIExtract@exe/ffsdumper.cpp.o' 'UEFIExtract/UEFIExtract@exe/uefidump.cpp.o' '-Wl,--no-undefined' '-Wl,--as-needed' 'common/liblzma.a' 'common/libbstrlib.a' 'common/libuefitoolcommon.a' '-lz'  
/usr/bin/ld: common/libuefitoolcommon.a(utility.cpp.o): in function `decompress(UByteArray const&, unsigned char, unsigned char&, unsigned int&, UByteArray&, UByteArray&)':
/tmp/UEFITool/bdir/../common/utility.cpp:240: undefined reference to `EfiTianoGetInfo'
/usr/bin/ld: /tmp/UEFITool/bdir/../common/utility.cpp:257: undefined reference to `TianoDecompress'
/usr/bin/ld: /tmp/UEFITool/bdir/../common/utility.cpp:259: undefined reference to `EfiDecompress'
/usr/bin/ld: /tmp/UEFITool/bdir/../common/utility.cpp:295: undefined reference to `LzmaGetInfo'
/usr/bin/ld: /tmp/UEFITool/bdir/../common/utility.cpp:298: undefined reference to `LzmaGetInfo'
/usr/bin/ld: /tmp/UEFITool/bdir/../common/utility.cpp:316: undefined reference to `LzmaDecompress'
/usr/bin/ld: /tmp/UEFITool/bdir/../common/utility.cpp:340: undefined reference to `LzmaGetInfo'
/usr/bin/ld: /tmp/UEFITool/bdir/../common/utility.cpp:352: undefined reference to `LzmaDecompress'
/usr/bin/ld: /tmp/UEFITool/bdir/../common/utility.cpp:365: undefined reference to `x86_Convert'
collect2: error: ld returned 1 exit status
```

After a bit of bisecting I determined that this project hits up against https://github.com/mesonbuild/meson/commit/ae9b23832e3ef4064e5735265ce7008794ab0491 which is a bug fixed in 0.41.0, the earliest version of Meson I successfully built UEFIExtract with.

eli-schwartz: TBH I think that's fine, Ubuntu 18.04 has Meson 0.45.0 anyway, and while Debian 9 originally shipped with Meson 0.37.1, it has 0.49.0 in stretch-backports.

UEFIExtract/meson.build
hughsie: TIL, fixing, thanks.

meson.build
hughsie: https://github.com/LongSoft/UEFITool/pull/285

PR #285, opened 2022-09-02
Meson more

hughsie:


meson.build
```diff
@@ -1,11 +1,12 @@
 project('UEFITool', ['c', 'cpp'],
   version: 'A60',
```
NikolajSchlej: A61

hughsie: Fixed.

NikolajSchlej:
LGTM, but we need to do something with that "version" field:
- if it's optional, I propose to just leave it unset for now
- if not, we need to change version.h format into something that doesn't require sed/awk to preprocess it, as we currently do to name releases.

NikolajSchlej:
I'll open a separate issue on that version thing, no need to stall this PR for it.

hughsie:
> if it's optional

It's kinda optional, etc:

    Distribution package /home/hughsie/Code/UEFITool/build/meson-dist/UEFITool-undefined.tar.xz tested
    Created /home/hughsie/Code/UEFITool/build/meson-dist/UEFITool-undefined.tar.xz

Do any of these seem acceptable? https://stackoverflow.com/questions/59201214/can-the-meson-project-version-be-assigned-dynamically

PR #287, opened 2022-09-08
Add GUIDs from Lenovo UEFI BIOS

crass:
These GUIDs name mappings were obtained by using the command `dh -v` from the Tianocore's UEFI shell, grepping lines with FvFile and extracting the shown name. This method should be able to be used for other firmwares.


NikolajSchlej:
Another method is to dump the FW, open the dumped binary with UEFITool NE, then do "File -> Export discovered GUIDs", and merge the resulting file with guids.csv.

I have nothing against the PR itself though, LGTM.

Issue #219, opened 2020-07-31
BootGuard parser under construction


hughsie:
Using the new_engine branch:

    terminate called after throwing an instance of 'std::out_of_range'
      what():  basic_string::substr: __pos (which is 18446744073698213888) > this->size() (which is 4194304)

    Program received signal SIGABRT, Aborted.
    0x00007ffff7ac59e5 in raise () from /lib64/libc.so.6
    (gdb) bt
    #0  0x00007ffff7ac59e5 in raise () from /lib64/libc.so.6
    #1  0x00007ffff7aae895 in abort () from /lib64/libc.so.6
    #2  0x00007ffff7e52961 in __gnu_cxx::__verbose_terminate_handler() [clone .cold] () from /lib64/libstdc++.so.6
    #3  0x00007ffff7e5e42c in __cxxabiv1::__terminate(void (*)()) () from /lib64/libstdc++.so.6
    #4  0x00007ffff7e5e497 in std::terminate() () from /lib64/libstdc++.so.6
    #5  0x00007ffff7e5e749 in __cxa_throw () from /lib64/libstdc++.so.6
    #6  0x00007ffff7e5544d in std::__throw_out_of_range_fmt(char const*, ...) [clone .cold] () from /lib64/libstdc++.so.6
    #7  0x00007ffff7efd9c8 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::substr(unsigned long, unsigned long) const () from /lib64/libstdc++.so.6
    #8  0x000000000042573d in UByteArray::mid(int, int) const ()
    #9  0x000000000042fae7 in FfsParser::checkProtectedRanges(UModelIndex const&) ()
    #10 0x0000000000432947 in FfsParser::performSecondPass(UModelIndex const&) ()
    #11 0x000000000043d594 in FfsParser::parse(UByteArray const&) ()
    #12 0x0000000000411f91 in main ()

When trying to extract this file: https://people.freedesktop.org/~hughsient/temp/BIOS_FV_BB.bin

Found on the LVFS... It could be the file isn't actually valid, but I'm pretty sure you don't want an exception.

NikolajSchlej:
Seriously need to harden that checkProtectedRanges() one, as we already had crashes in it, and the "fix" we had before is just silencing the crash.

vit9696:
Added a temporary workaround for now. Not closing as it needs thorough investigation. `addressDiff` looks apparently wrong for unclear reasons.

[...]

PR #288, opened 2022-09-09
Kaitai + BootGuard == <3

NikolajSchlej:
As the first step towards automated parsing, this change set replaces outdated BootGuard-related parsers with shiny new KaitaiStruct-based ones.

It also does the following:
- improves Intel FIT definitions by using the relevant specification
- adds sha1, sha384, sha512 and sm3 digest implementations
- updates LZMA SDK to v22.01
- moves GUIDs out of include files to prevent multiple instantiations
- enforces C++11
- adds Kaitai-based parsers for Intel FIT, BootGuard v1 and BootGuard v2 structures
- makes many small refactorings here, there and everywhere


NikolajSchlej:
Sorry for this change being a 15K+ lines of code, but I could not put it in unfinished. ;)

common/digest/sha512.c
```diff
@@ -0,0 +1,313 @@
...
+
+    if (md == NULL) return -1;
+    if (out == NULL) return -1;;
+
+    if (md->curlen >= sizeof(md->buf)) {
+       return -1;
+    }
+
+   sha512_done(md, buf);
+   memcpy(out, buf, 48);
```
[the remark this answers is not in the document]
NikolajSchlej: Well, duh, we only need 48 bytes of SHA384 hash, but also need 64 bytes buffer for `sha512_process` that is required to compute it. Dismissed.

common/generated/intel_acm.cpp
```diff
@@ -0,0 +1,92 @@
...
+    _clean_up();
+}
+
+void intel_acm_t::_clean_up() {
+    if (m_header) {
+        delete m_header; m_header = 0;
+    }
+}
+
+intel_acm_t::header_t::header_t(kaitai::kstream* p__io, intel_acm_t* p__parent, intel_acm_t* p__root) : kaitai::kstruct(p__io) {
```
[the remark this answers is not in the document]
NikolajSchlej:
```
    - id: rsa_exponent
      type: u4
      if: header_version == 0
```
This is why it can potentially be left uninitialised, and this is the relevant code bit that guarantees that we will not be reading that uninitialised value:
```
        // Add RsaPublicKey
        if (header->_is_null_rsa_exponent() == false) {
            acmInfo += usprintf("ACM RSA Public Key (Exponent: %Xh):", header->rsa_exponent());
        }
        else {
            acmInfo += usprintf("ACM RSA Public Key (Exponent: %Xh):", INTEL_ACM_HARDCODED_RSA_EXPONENT);
        }
```

common/digest/sha1.c
```diff
@@ -0,0 +1,251 @@
...
+       FF2(e,a,b,c,d,i++);
+       FF2(d,e,a,b,c,i++);
+       FF2(c,d,e,a,b,i++);
+       FF2(b,c,d,e,a,i++);
+    }
+
+    /* round four */
+    for (; i < 80; )  {
+       FF3(a,b,c,d,e,i++);
+       FF3(e,a,b,c,d,i++);
```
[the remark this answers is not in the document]
NikolajSchlej: Manual loop unrolling go BRRRRRR. Fixed in the next commit.

NikolajSchlej:
Forgot to mention, this is a fix for #219 among other things, @hughsie might be interested in trying it out.

NikolajSchlej:
Resolved an issue with incorrect constant that prevented FIT and everything in it from being found in any image. The lack of automated testing bites my rear once again...

UEFITool/CMakeLists.txt
```diff
@@ -13,6 +17,15 @@ SET(PROJECT_FORMS
  gotoaddressdialog.ui
  )
 
+SET(PROJECT_HEADERS
```
vit9696, line 20: Why does only UEFITool have headers now and not UEFIExtract/UEFIFind, where they were explicitly removed?

common/basetypes.h
```diff
@@ -203,7 +203,30 @@ typedef struct EFI_TIME_ {
...
-// SHA256 hash size in bytes
+// Hash sizes in bytes
+#define SHA1_HASH_SIZE   0x14
 #define SHA256_HASH_SIZE 0x20
+#define SHA384_HASH_SIZE 0x30
+#define SHA512_HASH_SIZE 0x40
+#define SM3_HASH_SIZE    0x20
+
+// TCG Algorithm Registry: Table 2
+#define HASH_ALGORITHM_SHA1    0x0004
```
vit9696: Add TCG_ prefix maybe?

common/digest/sha1.c
```diff
@@ -0,0 +1,230 @@
...
+
+    return 0;
+}
+
+void sha1(const void *in, unsigned long inlen, void* out)
+{
+    struct sha1_state ctx;
+    sha1_init(&ctx);
+    sha1_process(&ctx, (const unsigned char*)in, inlen);
+    sha1_done(&ctx, (unsigned char *)out);
```
vit9696, line 227: Not the case here, but I generally secure_memset the context after calculating the hash value. This prevents information leak on the stack. Perhaps better to change that here and in other hash implementations to avoid using the dangerous implementation where it matters.

common/digest/sha1.c
```diff
@@ -0,0 +1,230 @@
...
+   md->curlen = 0;
+   md->length = 0;
+   return 0;
+}
+
+static int sha1_process(struct sha1_state * md, const unsigned char *in, unsigned long inlen)
+{
+    unsigned long n;
+    int err;
+    if (md == NULL) return -1;
```
vit9696, line 141: These ifs are terrible, because the caller does not check for the return value. Please get rid of them everywhere (in sha1/sha2) and make init, process, and done return void.

UEFITool/CMakeLists.txt
NikolajSchlej: Turns out CMake is smart enough to derive all headers for non-Qt project, but not smart enough to derive a subset of Qt headers that require UIC pass to properly work, so some of those had to stay. The current list of headers is the minimal one, i.e. if you remove any, CMake builds will start failing.

common/basetypes.h
NikolajSchlej: Sure, why not.

common/digest/sha1.c
NikolajSchlej: We don't need to do it here because we do not perform any security-sensitive operations. The original code from LibTomCrypt already has that resolved, but I explicitly removed that part because we do not care, and, in case of UEFITool-based tools will never start caring.

common/digest/sha1.c
NikolajSchlej: To the contrary, the caller checks those values sometimes, see `*_process`, and I do not want to remove them because it will make porting upcoming changes from LibTomCrypt harder. All those hash implementation files got a "minimal set of changes to make them a) build, b) work, c) standalone, if possible" treatment, nothing else.

common/ffsparser.cpp
```diff
@@ -3366,156 +3379,233 @@ USTATUS FfsParser::checkProtectedRanges(const UModelIndex & index)
...
         bgProtectedRangeFound = false;
     }
     
     if (bgProtectedRangeFound) {
-        UByteArray digest(SHA256_DIGEST_SIZE, '\x00');
-        sha256(protectedParts.constData(), protectedParts.size(), digest.data());
+        UINT8 digest[SHA512_HASH_SIZE] = {};
+        UString digestString;
+        UString ibbDigests;
+        // SHA1
```
vit9696, line 3402: Maybe put that to a function and then call with hash pointer, ascii name and digest size? Feels a bit too copy-pasted.

NikolajSchlej: Yep, there's are lots of such in fitparser.cpp also.

common/fitparser.cpp
```diff
@@ -0,0 +1,1090 @@
+#include "fitparser.h"
```
vit9696, line 13: Copyright?

NikolajSchlej: Sure.

common/fitparser.cpp
```diff
@@ -0,0 +1,1090 @@
...
+    if (!fitIndex.isValid()) {
+        // Nothing to parse further
+        return U_SUCCESS;
+    }
+    // Explicitly set the item containing FIT as fixed
+    model->setFixed(fitIndex, true);
+    
+    // Special case of FIT header
+    UByteArray fitBody = model->body(fitIndex);
+    const INTEL_FIT_ENTRY* fitHeader = (const INTEL_FIT_ENTRY*)(fitBody.constData() + fitOffset);
```
vit9696, line 65: Where does this check for oob access?

common/fitparser.cpp
```diff
@@ -0,0 +1,1090 @@
...
+    
+    // Check FIT checksum, if present
+    UINT32 fitSize = fitHeader->Size * sizeof(INTEL_FIT_ENTRY);
```
vit9696, line 68: Can this overflow?

common/fitparser.cpp
```diff
@@ -0,0 +1,1090 @@
...
+            // Found it, no need to process further
+            return;
+        }
+    }
+    
+    // Check for all FIT signatures in item body
+    UByteArray lastVtfBody = model->body(ffsParser->lastVtf);
+    UINT64 fitSignatureValue = INTEL_FIT_SIGNATURE;
+    UByteArray fitSignature((const char*)&fitSignatureValue, sizeof(fitSignatureValue));
+    UINT32 storedFitAddress = *(const UINT32*)(lastVtfBody.constData() + lastVtfBody.size() - INTEL_FIT_POINTER_OFFSET);
```
vit9696, line 224: Is this guaranteed to produce valid address?

common/fitparser.cpp
```diff
@@ -0,0 +1,1090 @@
...
+    realSize = ucodeHeader->TotalSize;
+    
+    return U_SUCCESS;
+}
+
+USTATUS FitParser::parseFitEntryAcm(const UByteArray & acm, const UINT32 localOffset, const UModelIndex & parent, UString & info, UINT32 &realSize)
+{
+    // Parse Intel ACM using Kaitai-generated parser
+    // TODO: need to do something with the copy here, probably implement a custom wrapper
+    std::istringstream is(std::string(acm.constData(), acm.size()));
```
vit9696, line 283: Ugh, will std::string work with \0 bytes? You can use something like this: https://tuttlem.github.io/2014/08/18/getting-istream-to-work-off-a-byte-array.html

common/fitparser.cpp
```diff
@@ -0,0 +1,1090 @@
...
+                            header->error_entry_point(),
+                            header->gdt_max(),
+                            header->gdt_base(),
+                            header->segment_sel(),
+                            header->entry_point(),
+                            header->key_size() * (UINT32)sizeof(UINT32),
+                            header->scratch_space_size() * (UINT32)sizeof(UINT32));
+        
+        // Add RsaPublicKey
+        if (header->_is_null_rsa_exponent() == false) {
```
vit9696, line 365: !header->_is_null_rsa_exponent(), and better swap these two. Do if (header->_is_null_rsa_exponent()) first.

common/fitparser.h
```diff
@@ -0,0 +1,100 @@
...
+class FfsParser;
+
+#ifdef U_ENABLE_FIT_PARSING_SUPPORT
+class FitParser
+{
+public:
+    // Default constructor and destructor
+    FitParser(TreeModel* treeModel, FfsParser* parser) : model(treeModel), ffsParser(parser),
+        bgAcmFound(false), bgKeyManifestFound(false), bgBootPolicyFound(false) {}
+    ~FitParser() {}
```
vit9696, line 37: Do not defined destructor if you do not need it.

common/fitparser.h
```diff
@@ -0,0 +1,100 @@
...
+    std::vector<std::pair<std::vector<UString>, UModelIndex> > getFitTable() const { return fitTable; }
+    
+    // Obtain security info
+    UString getSecurityInfo() const { return securityInfo; }
+    
+    // FIT parsing
+    USTATUS parseFit(const UModelIndex & index);
+        
+private:
+    TreeModel *model;
```
vit9696, line 54: To be honest, maybe zero these inline for safety? Like TreeModel * model {nullptr}; and so on?

common/kaitai/custom_decoder.h
```diff
@@ -0,0 +1,16 @@
+#ifndef KAITAI_CUSTOM_DECODER_H
+#define KAITAI_CUSTOM_DECODER_H
+
+#include <string>
+
+namespace kaitai {
+
+class custom_decoder {
```
vit9696, line 8: Is this used? I think this was the draft for the istream thing I posted earlier.

common/fitparser.cpp
NikolajSchlej: Added to findFitRecursive, thanks for catching it.

common/fitparser.cpp
NikolajSchlej: Not in sense of arithmetic, but in sense of OBB, check added to parseFit, thanks for catching.

common/meparser.cpp
```diff
@@ -54,7 +53,7 @@ USTATUS MeParser::parseMeRegionBody(const UModelIndex & index)
     
     // Check ME signature to determine it's version
     // ME v11 and older layout
-    if (meRegion.left(sizeof(UINT32)) == FPT_HEADER_SIGNATURE || meRegion.mid(ME_ROM_BYPASS_VECTOR_SIZE, sizeof(UINT32)) == FPT_HEADER_SIGNATURE) {
+    if (*(UINT32*)meRegion.constData() == FPT_HEADER_SIGNATURE || *(UINT32*)(meRegion.constData() + ME_ROM_BYPASS_VECTOR_SIZE)  == FPT_HEADER_SIGNATURE) {
```
vit9696, line 56: Is this guaranteed to have 4 bytes?

common/meparser.cpp
```diff
@@ -73,7 +72,7 @@ USTATUS MeParser::parseMeRegionBody(const UModelIndex & index)
         return U_INVALID_ME_PARTITION_TABLE;
     }
     // Data partition always points to FPT header
-    if (meRegion.mid(ifwi16Header->DataPartition.Offset, sizeof(UINT32)) == FPT_HEADER_SIGNATURE) {
+    if (*(UINT32*)(meRegion.constData() + ifwi16Header->DataPartition.Offset) == FPT_HEADER_SIGNATURE) {
```
vit9696, line 75: Same here and below.

kaitai_regenerate.sh
```diff
@@ -0,0 +1,41 @@
...
+elif [ "${UTARGET/MINGW32/}" != "$UTARGET" ]; then
+  export UPLATFORM="win32"
+else
+  # Fallback to something...
+  export UPLATFORM="$UTARGET"
+fi
+
+# Generate
+echo "Attempting to to generate parsers from Kaitai KSY files on ${UPLATFORM}..."
+kaitai-struct-compiler --target cpp_stl --outdir common/generated common/ksy/*
```
vit9696: Add || exit 1 in my opinion.

common/fitparser.cpp
NikolajSchlej: Yes, as the minimal size of UEFI image (volume header + ffs file header) is more than 0x40 (the value of INTEL_FIT_POINTER_OFFSET).

common/fitparser.cpp
NikolajSchlej: This is from Kaitai docs, and it explicitly does work with \0 bytes.
Screenshot 2022-09-10 at 11 39 45

common/fitparser.cpp
NikolajSchlej: I do not like it, because in `!header` is invisible, as `!` and `h` are both `|`-shaped. 
Will probably swap the order to be `false == ...` to prevent occasional ==/= confusion.

common/fitparser.h
NikolajSchlej: Copied from other places, needs to be removed everywhere.

common/fitparser.h
NikolajSchlej: They are initialised by constructor init list, and if I also do it here, analysers will berate me for it. SonarCloud already does. :)

common/kaitai/custom_decoder.h
NikolajSchlej: No, the whole folder is a verbatim copy of https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/tree/master/kaitai
It can start being used in the future, if we will be using custom decoders.

common/meparser.cpp
NikolajSchlej: Yes, but it needs to be replaced with readUnaligned, because all this `*(UINTX)constCharPtr` shit breaks strict aliasing anyway. I will make a round of refactoring specifically addressing that later.

kaitai_regenerate.sh
NikolajSchlej: Sure.

common/meparser.cpp
NikolajSchlej: Same as above, needs to be addressed in another PR, let's make an issue for it.

common/fitparser.h
vit9696: Now that we have C++11, perhaps it is the time to get rid of constructor lists? They are legacy and fugly. Probably better in a separate commit.

NikolajSchlej: Separate commit for sure.

PR #293, opened 2022-09-10
Make Boot Guard protected/unprotected/partially protected ranges readable in dark mode

yeggor:
Hi. In dark mode, the colors chosen to mark protected/unprotected/partially protected ranges make the information unreadable. Below are a few screenshots:
default1
default2

I propose to modify the colors a bit if user will rebuild the `UEFITool` by specifying the optional parameter `-DDARK_MODE=ON` with cmake.

In result it will look better:
dark1
dark2


NikolajSchlej:
The colors are really nice, but requiring two separate builds for light and dark modes is a no-go. I'll go read about how to detect the current mode on both macOS and Windows, and how to apply the colors in a way that doesn't require changes in ffsparser.cpp, only in uefitool.cpp.

yeggor:
Unquestionably. It would be great if it worked like you said.
I'm closing this PR then

PR #296, opened 2022-09-14
Fix build on FreeBSD

pkubaj:
Provide correct macros for byte swapping on FreeBSD.


NikolajSchlej:
LGTM, though it only fixes FreeBSD, and there are some other BSD variants that will remain unsupported, but most likely use the same exact functions: `__OpenBSD__`, `__NetBSD__`, `__DragonFly__`. 
Could you please add the other ones too? In return I promise to install each in a VM and test the result.

pkubaj:
Since you merged it, I consider your request cancelled.

NikolajSchlej:
Indeed, made the required changes myself after the merge, because the VMs were ready.

PR #297, opened 2022-09-17
Compilation fixes

ValdikSS:


NikolajSchlej:
@ValdikSS, thank you for the PR, appreciated. Could you please add more information on what environment exactly is this fix for? I can derive Fedora from the commit, but still need the version to be able to test it.

ValdikSS:
Well, `qmake-qt5` is because there's no `qmake` on Fedora, the other commit is to utilize parallel compilation on multiple cores (getconf is cross-platform), and the other commit is a fix in readme.

I tested on Fedora 35.

PR #316, opened 2022-10-27
Fix possible while true while parsing FV headers

mefff:
While parsing the EFI_FV_BLOCK_MAP_ENTRY array in the FV header it's possible to get into a while true.

I couldn't get much info about EFI_FV_BLOCK_MAP_ENTRY, don't know why it's there since there is already a size. So another possible solution would be to skip the header all together if we fail to parse the EFI_FV_BLOCK_MAP_ENTRY array.


NikolajSchlej:
Do you have a file that reproduces this issue?

mefff:
No, just found it digging into the code

NikolajSchlej:
That code feels like a broken overflow check done by (ab)using pointer arithmetics. I doubt it ever fired because such an UEFI image would not be bootable (if we can't parse it, UEFI also can not).
It's clear that "continue" is not what needs to be done there, instead, if we have a block map so large that it overflows the volume, the whole volume needs to be marked as broken and unparsable. 
Let's convert this into an issue, and I'll fix it properly later.

NikolajSchlej:
#317

PR #331, opened 2023-01-10
Add support for PROTECTED_RANGE_VENDOR_HASH_AMI_V3

cybojanek:


common/ffsparser.h
```diff
@@ -57,6 +57,11 @@ typedef struct PROTECTED_RANGE_ {
     UINT8      Type;
     UINT8      : 8;
     UByteArray Hash;
+
+    // Additional offsets and sizes.
```
cybojanek, line 61: This is not a pretty extension of the struct, but it prevented a lot of other refactoring, because all the other code assumes  `PROTECTED_RANGE` has only one offset and size.

If you have an alternate suggestion for `PROTECTED_RANGE`, I can try to incorporate it and refactor the rest of the code.

NikolajSchlej:
Thanks a ton for the PR, needs some cleanup before it can be merged, I'll comment inline.

common/ffs.h
```diff
@@ -828,11 +828,21 @@ typedef struct PROTECTED_RANGE_VENDOR_HASH_FILE_HEADER_AMI_V2_
     BG_VENDOR_HASH_FILE_ENTRY Hash1;
 } PROTECTED_RANGE_VENDOR_HASH_FILE_HEADER_AMI_V2;
 
+// NOTE(cybojanek):
```
NikolajSchlej: Line 831 needs to be removed.

common/ffsparser.cpp
```diff
@@ -3844,7 +3890,61 @@ USTATUS FfsParser::parseVendorHashFile(const UByteArray & fileGuid, const UModel
...
+                securityInfo += "\nPadding:";
+                for (size_t i = 0; i < 12; i++) {
+                    securityInfo += usprintf(" %08Xh", entry->Padding[i]);
+                    if (entry->Padding[i] != 0xFFFFFFFF) {
+                        unknownPadding = true;
+                    }
+                }
+                securityInfo += "\n\n";
+
+                // NOTE(cybojanek): The offset and size calculations seem to
```
NikolajSchlej: "NOTE(cybojanek):" needs to be removed.

common/ffsparser.h
NikolajSchlej: Yeah, this is definitely somewhat dirty, let's see if we can invent a better solution. Mind sharing a test image?

NikolajSchlej:
@cybojanek, bump. Still want to see the file you've used for testing this change set.

common/ffsparser.h
cybojanek: Sorry for losing track of this. The bump was very welcome.

Sample image can be downloaded from: https://www.supermicro.com/en/support/resources/downloadcenter/firmware/MBD-X12SDV-10C-SP6F/BIOS

NikolajSchlej:
Using this one as a source of inspiration to a slightly different implementation that I'm writing now. Will close the PR as is, but mention you as the original author.

NikolajSchlej:
Done in https://github.com/LongSoft/UEFITool/commit/33c25e8255b3f5568b2c528bf629b21b625b99b2

cybojanek:
Awesome!

Thanks for UEFITool :-)

PR #353, opened 2023-03-16
Fix FfsParser issues found by fuzzing (2)

yeggor:
Hi @NikolajSchlej.
Using the fuzzer that was recently added, I caught all the crashes you originally tried to fix here https://github.com/LongSoft/UEFITool/commit/b8567d32cc158eb68d900d9a161e92889e643627 (and more).

After all these fixes, so far I have not been able to catch any more crashes with default configuration.


NikolajSchlej:
Lets hope this won't make any image regress...

PR #370, opened 2023-07-07
Display message when search fails

druckdev:
Display a message 'could not be found' when the UEFITool search fails.

### DIFF
--- a/UEFITool/ffsfinder.cpp
+++ b/UEFITool/ffsfinder.cpp
@@ -19,6 +19,14 @@
 #include <QRegExp>
 #endif
 
+USTATUS FfsFinder::findHexPattern(const UByteArray & hexPattern, const UINT8 mode) {
+    const UModelIndex rootIndex = model->index(0, 0);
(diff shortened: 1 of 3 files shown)


druckdev:
I was quite unsure on how to best format some of the long lines. Please do tell me if there is a Coding Style Guide or similar.

NikolajSchlej:
There is no defined coding style, format as you see fit.

This design is somewhat strange though, as it exposes the msg() function of ffsFinder for unclear reason. UEFITool class should not care about messages generated by ffsFinder, it just relays them to Search panel.

druckdev:
I agree!

Could you please take a look at the new commit? If you don't like the wrapper function we could leave the old ones and just check at the end if `(ret != U_SUCCESS && index == mode->index(0,0))` but that seems less efficient.

druckdev:
(Renamed index back to rootIndex in the wrapper functions)

NikolajSchlej:
Wrappers are just fine, LGTM.

PR #373, opened 2023-08-03
Fix for AMI NVAR parser based on kaitai struct

chunqing286:
modify 1 line in ami_nvar.ksy (line 128).

Bug scenario:
nvar_entry.attributes.valid = 0
nvar_entry.attributes.extended_header = 1

Now, extended_header_size_field 's value is undefined, as defined in ami_nvar.ksy
```
extended_header_size_field:
     pos: _io.pos - sizeof<u2>
     type: u2
     if: _parent.attributes.valid
      and _parent.attributes.extended_header
      and _parent.size > sizeof<u4> + sizeof<u2> + sizeof<u4> + sizeof<u2>
```
but this value would be used by other instances, e.g. extended_header_size, as defined in ami_nvar.ksy
```
extended_header_size:
     value: '_parent.attributes.extended_header ? (extended_header_size_field >= sizeof<nvar_extended_attributes> + sizeof<u2> ? extended_header_size_field : 0) : 0'
```
finally leads to parse error/exception or program crash.

### DIFF
--- a/common/generated/ami_nvar.cpp
+++ b/common/generated/ami_nvar.cpp
@@ -6,24 +6,18 @@
 ami_nvar_t::ami_nvar_t(kaitai::kstream* p__io, kaitai::kstruct* p__parent, ami_nvar_t* p__root) : kaitai::kstruct(p__io) {
     m__parent = p__parent;
     m__root = this;
-    m_entries = 0;
-
-    try {
-        _read();
-    } catch(...) {
-        _clean_up();
-        throw;
-    }
+    m_entries = nullptr;
+    _read();
 }
 
 void ami_nvar_t::_read() {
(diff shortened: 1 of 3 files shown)


NikolajSchlej:
Thanks for the fix, makes sense. Need to regenerate the files by calling the most recent kaitai_regenerate.sh, because the current ones seem to be generated without C++11 additions (`kaitai-struct-compiler --target cpp_stl --cpp-standard 11`).

chunqing286:
Hi, thanks for your instructions about C++ 11 standard. 

I have regenerate the C++ header and source file by  `kaitai-struct-compiler --target cpp_stl --cpp-standard 11`. 

By the way , I use the lastest Kaitai Struct release version 0.10.

PR #374, opened 2023-08-17
Fix CPD Manifest's partition offset

XutaxKamay:
The header size needs to be probably added here, maybe it was forgotten ?

This is needed in order to get the proper base address of a CPD_EXT_IFWI_PARTITION_MANIFEST structure in the image for example in that specific case.

If I don't do this, it points to a meaningless address.

Let me know if there can be any issues with this.

### DIFF
--- a/common/ffsparser.cpp
+++ b/common/ffsparser.cpp
@@ -4673,7 +4673,7 @@ make_partition_table_consistent:
                                         manifestHeader->ExponentSize * (UINT32)sizeof(UINT32), manifestHeader->ExponentSize * (UINT32)sizeof(UINT32));
                         
                         // Add tree item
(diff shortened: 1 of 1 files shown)


XutaxKamay:
Sorry, but I just noticed that it does indeed make a problem, FTPR.man does not point anymore to the right base address, I think the header size will likely be needed to be added on the offset field in the Types::CpdExtension. I'll make another pull request to revert this and fix it again

NikolajSchlej:
For future similar changes let's start with a GH issue with files and screenshots demonstrating the problem, will be easier to prevent multiple fixes for the same problem this way.

XutaxKamay:
> For future similar changes let's start with a GH issue with files and screenshots demonstrating the problem, will be easier to prevent multiple fixes for the same problem this way.

I'll do that, thank you.

PR #389, opened 2024-02-20
Define ACCESSPERMS When Compiling With musl

mischif:
Compiling UEFITool fails on Linux distros using the musl libc, such as Alpine, due to ACCESSPERMS not being defined in `sys/stat.h`:

```
uefitool/common/filesystem.cpp: In function 'bool makeDirectory(const Bstrlib::CBString&)':
uefitool/common/filesystem.cpp:86:38: error: 'ACCESSPERMS' was not declared in this scope
   86 |     return (mkdir(dir.toLocal8Bit(), ACCESSPERMS) == 0);
      |                                      ^~~~~~~~~~~
```

This PR creates a definition for ACCESSPERMS, allowing compilation to complete.

### DIFF
--- a/common/filesystem.cpp
+++ b/common/filesystem.cpp
@@ -75,6 +75,9 @@ UString getAbsPath(const UString & path)
 #else
 #include <unistd.h>
 #include <stdlib.h>
+#if !defined(ACCESSPERMS)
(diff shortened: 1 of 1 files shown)


NikolajSchlej:
Thanks, need to add such a linux into CI/CD matrix, so we can catch stuff like this next time

PR #399, opened 2024-07-01
Set desktop file name to fix the missing icon when running under Wayland

Integral-Tech:


NikolajSchlej (approved):

NikolajSchlej:
"D:\a\UEFITool\build\UEFITool\UEFITool.vcxproj" (Rebuild target) (1) ->
(ClCompile target) -> 
  d:\a\uefitool\uefitool\uefitool\uefitool_main.cpp(29): error C3861: 'setDesktopFileName': identifier not found [D:\a\UEFITool\build\UEFITool\UEFITool.vcxproj]

Now available in Qt 5.6.3 we still support, needs an ifdef/endif guards around to not fail compilation.

"This property was introduced in Qt 5.7."

Integral-Tech:
> "D:\a\UEFITool\build\UEFITool\UEFITool.vcxproj" (Rebuild target) (1) -> (ClCompile target) -> d:\a\uefitool\uefitool\uefitool\uefitool_main.cpp(29): error C3861: 'setDesktopFileName': identifier not found [D:\a\UEFITool\build\UEFITool\UEFITool.vcxproj]
> 
> Now available in Qt 5.6.3 we still support, needs an ifdef/endif guards around to not fail compilation.
> 
> "This property was introduced in Qt 5.7."

Fixed in my latest commit :)

NikolajSchlej:
No need for Q_OS_LINUX, as the property is supported everywhere and we do support FreeBSD and other BSD variants that do use Gnome and/or KDE, so setting the desktop file name for them makes sense.

Integral-Tech:
> No need for Q_OS_LINUX, as the property is supported everywhere and we do support FreeBSD and other BSD variants that do use Gnome and/or KDE, so setting the desktop file name for them makes sense.

Fixed :)

PR #400, opened 2024-07-03
perf: use more efficient QString constructors

Integral-Tech:


NikolajSchlej:
I don't think any measurable performance gain can be achieve by making this changes, and they introduce a concept of QLatin1String that had not been used in the codebase before. I'm all for cleanup and refactoring, but this kind doesn't look good in my book.

@vit9696, what do you think?

Integral-Tech:
> I don't think any measurable performance gain can be achieve by making this changes, and they introduce a concept of QLatin1String that had not been used in the codebase before. I'm all for cleanup and refactoring, but this kind doesn't look good in my book.
> 
> @vit9696, what do you think?

[Qt Weekly](https://www.qt.io/blog/2014/06/13/qt-weekly-13-qstringliteral) suggests using ```QLatin1String``` & ```QStringLiteral``` over ```QString```

NikolajSchlej:
It is fine as suggestions go, but in the grand scheme of things, this change is made for the sake of itself, not performance. Any performance updates need to be coupled with two sets of measurements - "before" and "after", otherwise calling them "performance improvements" makes very little sense. In this PR, neither it was measurably slower before, not it becomes measurably faster after, so I will decline it as a NOOP, despite the advice from Qt Weekly. We used only QString in Qt-enabled portion of our codebase, and we will continue to use it, because it's much easier to reason about strings this way.

vit9696:
@NikolajSchlej, a bit late to comment, but I agree with your judgement. The code becomes less intuitive with mixed use of QString and QLatin1String, and it is unjustified that the changed code is on hot path and needs performance adjustments. Perhaps if @Integral-Tech cares about performance, he could perform some evaluation on that matter and files an issue describing the results based on actual user experience and tooling analysis.

PR #412, opened 2025-03-08
Replace most NVRAM parsers with ones generated from KaitaiStruct definitions

NikolajSchlej:
Back in 2023 the initial attempt to fuzz the parsing engine with LibFuzzer and HongFuzz quickly failed because the fuzzer found an issue with our NVRAM parsers requiring backtracking to parse things properly, and that effectively dead-looped them on way too many possible inputs to count.

As an effort to get rid of that, and rewrite manual parsers on something more declarative and saner, I've started to rewrite NVRAM parsers in KaitaiStruct, and the first one (AMI NVAR) had been in use since about A62, and proven to work better than expected. Then I had to return to the US and restart my full-day employment, so all work on UEFITool had to be postponed. Until this month that is, and this PR is the result of it.

@vit9696, please review, as I do not expect this new code to change drastically any time soon. I will be doing cross-tests between A69 and this branch, trying to catch any possible offset errors and other things that Kaitai's codegen can not cover.


NikolajSchlej:
Will have to make a bit more changes here for the sake of optimization, turns out trying every parser on every byte got rather slow if it creates a ton of C++ objects and destroys those objects immediately.

Update: done, ready for review.

NikolajSchlej:
Local cross-testing with about 30 different samples of Insyde H2O and Phoenix SCT based FWs (AMI NVAR parser is not affected by this change) showed no unexpected results. 
Several bugs present in A69 got fixed by this PR:
- EDK2 FTW header had the CRC value reset to 0xFFFFFFFF, now is getting parsed properly.
- Insyde Factory Data Copy store had wrong offset (+8 from the real one), now fixed.

common/ffsparser.cpp
```diff
@@ -1415,6 +1530,28 @@ continue_searching: {}
             nextItemOffset = offset;
             break;
         }
+        else if (readUnaligned(currentPos) == INSYDE_FLASH_DEVICE_MAP_SIGNATURE) {
+            // Check data size
+            if (restSize < sizeof(INSYDE_FLASH_DEVICE_MAP_HEADER))
+                continue;
+            
+            const INSYDE_FLASH_DEVICE_MAP_HEADER *fdmHeader = (const INSYDE_FLASH_DEVICE_MAP_HEADER *)currentPos;
```
vit9696, line 1538: This can be misaligned.

common/ffsparser.cpp
```diff
@@ -948,6 +952,114 @@ USTATUS FfsParser::parseRawArea(const UModelIndex & index)
...
+                                    parsed.entry_size(), parsed.entry_size(),
+                                    parsed.entry_format(),
+                                    parsed.revision(),
+                                    parsed.num_extensions(),
+                                    (UINT32)parsed.fd_base_address(),
+                                    parsed.checksum());
+            
+            // Check header checksum
+            {
+                UByteArray tempHeader = data.mid(itemOffset, sizeof(INSYDE_FLASH_DEVICE_MAP_HEADER));
```
vit9696, line 982: Why cannot this throw and if it can, where is the exception handler? I do not see any from uefiextract_main.cpp to this function, maybe blind though.

common/nvramparser.cpp
```diff
@@ -39,6 +52,14 @@ USTATUS NvramParser::parseNvarStore(const UModelIndex & index)
     if (nvar.isEmpty())
         return U_SUCCESS;
 
+    // Obtain required fields from parsing data
+    UINT8 emptyByte = 0xFF;
+    if (model->hasEmptyParsingData(index) == false) {
+        UByteArray data = model->parsingData(index);
+        const VOLUME_PARSING_DATA* pdata = (const VOLUME_PARSING_DATA*)data.constData();
```
vit9696, line 59: I do not quite get, why are we guaranteed that our parent is a volume? Because if it is not, VOLUME_PARSING_DATA may be out of bounds.

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
-        // Padding between stores
-        if (storeOffset > prevStoreOffset + prevStoreSize) {
-            UINT32 paddingOffset = prevStoreOffset + prevStoreSize;
-            UINT32 paddingSize = storeOffset - paddingOffset;
-            UByteArray padding = data.mid(paddingOffset, paddingSize);
+        // VSS
+        try {
+            if (volumeBodySize - storeOffset < sizeof(VSS_VARIABLE_STORE_HEADER)) {
+                // No need to parse further, the rest of the volume is too small
+                goto not_vss;
```
vit9696, line 338: Why not throw domain_error or out_of_range instead of goto here and below? Feels more readable this way.

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
+            }
             
-            // Get info
-            name = UString("Padding");
-            info = usprintf("Full size: %Xh (%u)", (UINT32)padding.size(), (UINT32)padding.size());
+            // Perform initial sanity check
+            const VSS_VARIABLE_STORE_HEADER* storeHeader = (const VSS_VARIABLE_STORE_HEADER*)(volumeBody.constData() + storeOffset);
```
vit9696, line 342: You can get alignment issues here as well, also below in similar casts.

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
+                
+                const UINT32 variableAttributes = variable->attributes()->non_volatile()
+                + (variable->attributes()->boot_service() << 1)
+                + (variable->attributes()->runtime() << 2)
+                + (variable->attributes()->hw_error_record() << 3)
+                + (variable->attributes()->auth_write() << 4)
+                + (variable->attributes()->time_based_auth() << 5)
+                + (variable->attributes()->append_write() << 6)
+                + (UINT32)(variable->attributes()->reserved() << 7)
+                + (variable->attributes()->apple_data_checksum() << 31);
```
vit9696: bool will propagate to int, which is 32-bit on supported targets. 1 << 31 will change sign. Something is wonky here.

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
+                model->addItem(entryOffset, Types::VssEntry, subtype, name, text, info, header, body, UByteArray(), Fixed, headerIndex);
+                
+                entryOffset += variableSize;
+            }
+            
+            storeOffset += storeSize - 1;
+            previousStoreEndOffset = storeOffset + 1;
+            continue;
+        } catch (...) {
+           // Parsing failed, try something else
```
vit9696, line 532: When failing to parse in the middle of the item (e.g. when `mid` fails) I would expect us to guess the NVRAM type correctly, but have something corrupted (e.g. bad SPI flasher I/O). It is bad to simply retry with another parser without a single notification of what happened. I believe we should separate "corrupted data" exceptions from "different NVRAM type" exceptions. There should be a message for the first kind at least for debugging purposes.

The comment applies to the code below as well.

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
+not_vss:
+        // VSS2
+        try {
+            if (volumeBodySize - storeOffset < sizeof(VSS2_VARIABLE_STORE_HEADER)) {
+                // No need to parse further, the rest of the volume is too small
+                goto not_vss2;
+            }
+            
+            // Perform initial sanity check
+            const VSS2_VARIABLE_STORE_HEADER* storeHeader = (const VSS2_VARIABLE_STORE_HEADER*)(volumeBody.constData() + storeOffset);
```
vit9696, line 543: Alignment issues below. To think of it, might be worth using packed or something for safety. Since even modern RISC-V CPUs often raise an exception on unaligned access I believe we should not completely ignore the issue.

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
-    UINT32 storeOffset = prevStoreOffset;
-    UINT32 prevStoreSize = 0;
-    
-    while (!result) {
-        // Padding between stores
-        if (storeOffset > prevStoreOffset + prevStoreSize) {
-            UINT32 paddingOffset = prevStoreOffset + prevStoreSize;
-            UINT32 paddingSize = storeOffset - paddingOffset;
-            UByteArray padding = data.mid(paddingOffset, paddingSize);
+        // VSS
```
vit9696, line 334: This really badly needs to be split into functions. Currently this is extremely unreadable and has very unclear variable semantics.

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
+            if (guid != NVRAM_MAIN_STORE_VOLUME_GUID
+                && guid != EDKII_WORKING_BLOCK_SIGNATURE_GUID
+                && guid != VSS2_WORKING_BLOCK_SIGNATURE_GUID) {
+                // No need to parse further, not a FTW store
+                goto not_ftw;
+            }
+            // Determine store size
+            UINT32 storeSize;
+            if (storeHeader->WriteQueueSize % 0x10 == 4) {
+                storeSize = sizeof(EFI_FAULT_TOLERANT_WORKING_BLOCK_HEADER32) + storeHeader->WriteQueueSize;
```
vit9696, line 734: Why writeQueueSize arithmetics cannot overflow or at least lead to OOB here and below?

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
+        // Phoenix Flash Map
+        try {
+            if (volumeBodySize - storeOffset < sizeof(PHOENIX_FLASH_MAP_HEADER)) {
+                // No need to parse further, the rest of the volume is too small
+                goto not_flm;
+            }
+            // Perform initial sanity check
+            const PHOENIX_FLASH_MAP_HEADER* storeHeader = (const PHOENIX_FLASH_MAP_HEADER*)(volumeBody.constData() + storeOffset);
+            if (UByteArray((const char*)storeHeader->Signature, NVRAM_PHOENIX_FLASH_MAP_SIGNATURE_LENGTH) != NVRAM_PHOENIX_FLASH_MAP_SIGNATURE
+                || storeHeader->NumEntries > 113) {
```
vit9696: Could you please at least add the meaning of 113 in a comment?

common/nvramparser.cpp
```diff
@@ -296,1548 +316,1225 @@ USTATUS NvramParser::parseNvramVolumeBody(const UModelIndex & index)
...
+                            // Add tree item
+                            model->addItem(entryOffset, Types::Padding, getPaddingType(freeSpace), UString("Padding"), UString(), info, UByteArray(), freeSpace, UByteArray(), Fixed, headerIndex);
+                        }
+                    }
                     break;
                 }
-            if (!reservedBytesValid)
-                continue;
+                
+                const EVSA_ENTRY_HEADER* entryHeader = (const EVSA_ENTRY_HEADER*)(volumeBody.constData() + storeOffset + entryOffset);
```
vit9696, line 1135: I might be missing the model here, but why is it safe to expect an object of EVSA_ENTRY_HEADER size at entryOffset?

vit9696 (commented):
Made a brief review. I mostly focused on useability and safety issues, trusting you and kaitai on the format matter.

common/ffsparser.cpp
NikolajSchlej: True, will add readUnaligned here.

common/ffsparser.cpp
NikolajSchlej: We have covered that check in findNextRawAreaItem:
```
        else if (readUnaligned(currentPos) == INSYDE_FLASH_DEVICE_MAP_SIGNATURE) {
            // Check data size
            if (restSize < sizeof(INSYDE_FLASH_DEVICE_MAP_HEADER))
                continue;
            
            const INSYDE_FLASH_DEVICE_MAP_HEADER *fdmHeader = (const INSYDE_FLASH_DEVICE_MAP_HEADER *)currentPos;
            
            if (restSize < fdmHeader->Size)
                continue;
```
No mid() calls will throw as they are operating on valid memory.

common/nvramparser.cpp
NikolajSchlej: Our parsing function is called 'parseNvramVolumeBody', and is called like this:
```
    // Parse NVRAM volume with a dedicated function
    if (model->subtype(index) == Subtypes::NvramVolume) {
        return nvramParser->parseNvramVolumeBody(index);
    }
```

common/nvramparser.cpp
NikolajSchlej: Sure, the map is of 0x1000 size, and 113 is the amount of entries that can still fit in. There's a comment about it in nvram.h, but it needs to be converted into a define.

common/nvramparser.cpp
NikolajSchlej: Slower, tried that.

common/nvramparser.cpp
NikolajSchlej: True, will be covered together with the other case.

common/nvramparser.cpp
NikolajSchlej: Proven to work, but I'll convert to UINT32 explicitly just in case.

common/nvramparser.cpp
NikolajSchlej: Yep, same.

common/nvramparser.cpp
NikolajSchlej: It can, sure, but it will be fixed by the following 'storeSize = MIN(volumeBodySize - storeOffset, storeSize);' and can only lead to parsing less stuff than we could otherwise.

common/nvramparser.cpp
NikolajSchlej: This is guaranteed by KaitaiStream source

common/nvramparser.cpp
NikolajSchlej: No, I refuse to add functions for things that don't need to be functions, never get called from anywhere else and not ever repeat outside of this one.

common/nvramparser.cpp
NikolajSchlej: This is a follow-up nice to have feature.

PR #419, opened 2025-03-19
Remove PATH_MAX from realpath

yangfl:
POSIX.1-2008 ensures realpath() give a dynamically allocated buffer if NULL is passed, which avoids using PATH_MAX, which may be too large to fit in stack, or even undefined on some systems.


NikolajSchlej:
Thanks, merged.

PR #438, opened 2025-06-23
Refactored UI with docks, added hex view dock, small menus fixes - cl…

Ykidia:
…ear files list and expand/collapse tree nodes

I would like to suggest you some changes to the interface:
- replaced fixed UI with docks;
- added hex view dock with header (if exists) highlighting;
- added search dialog settings save and restore;
- small fixes to "clear recently opened files list" and "Expand whole section"/"Collapse whole section" menu events as per https://github.com/LongSoft/UEFITool/issues/437

~~! There is a small problem still not fixed with docks - new UI looks like garbage if settings exist after previous UEFITool with classic UI launches. Before running UEFITool with docks please delete those settings.~~

! In Wayland, docks will not float.

! Because of new hex editing features I've added (but still not published) using existing menu events "hex view"/"hex body view", I don't know, what to do with existing classic "hex view"/"hex body view"; but they are still usable even if hex view dock is visible.

Enjoy!

### DIFF
--- a/UEFITool/uefitool.cpp
+++ b/UEFITool/uefitool.cpp
@@ -1241,10 +1241,11 @@ void UEFITool::contextMenuEvent(QContextMenuEvent* event)
 
     if (menu) {
         QList<QAction*> actions = menu->actions();
-        QAction s = QAction(nullptr);
-        s.setSeparator(true);
+        QAction *separator = new QAction(nullptr);
+        separator->setSeparator(true);
         QMenu::exec(
-            actions << &s << ui->actionExpandRecursively << ui->actionCollapseRecursively, gp);
+            actions << separator << ui->actionExpandRecursively << ui->actionCollapseRecursively, gp);
+        delete separator;
     }
 }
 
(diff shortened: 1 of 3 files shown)


Ykidia:
Now all fixed, that's how it looks like:

UEFITool-docks1

UEFITool-docks2

UEFITool-docks3

NikolajSchlej:
Thanks, looks interesting from the source PoV, will check it out a bit later when life gets less hectic.

NikolajSchlej:
@Ykidia, the change set looks fine, but now Debug version get EXC_BAD_ACCESS in UEFITool class destructor. I'll try to debug it, but you might be more familiar with your code to see where the stuff could be deleted twice.

NikolajSchlej:
Reverted for now, will return to it after further discussion, please open an Issue where we all (@vit9696 in particular) can discuss it.

Ykidia:
> now Debug version get EXC_BAD_ACCESS in UEFITool class destructor

I suppose this is QHexView, I got similar behavior when using QHexView for editing and solved this by setData(QByteArray()) before every new data set.
But now the problem you mentioned is not reproducible for me, neither in Windows, nor in Ubuntu (Wayland).

Ykidia:
> please open an Issue where we all (@vit9696 in particular) can discuss it

How exactly should I do this? Suggest changing the interface towards docks, because it will allow doing this and that, for example, hiding the not used builder tab (and others), showing the hex preview where convenient (and others) and so on?

NikolajSchlej:
Created one for you, let's use it for further discussion: https://github.com/LongSoft/UEFITool/issues/439

PR #443, opened 2025-09-15
AMD-specific structures parsing

Ykidia:
* added AMD Embedded Firmware Structure parsing;
* fixed jump to tree view item when clicked in parser info window - when inserting items not only at end (for example, CREATE_MODE_BEFORE), QModelIndex::row() of all superior items at same level becomes invalid and needed to be updated;
* added correct CPU addresses mapping if more than single bank detected, checking that CPU addresses adding only for UEFI image or BIOS region items info;
* added QComboBox to gotoAddressDialog with banks list (if more than single), to specify in which bank to search and jump.

Tested compiling and working on Windows, Ubuntu 25.04, macOS Sequoia.

TODO:
* display EFT(s) as in FIT: need more time to decide how it should look (replace table in FIT and rename FIT? add another dock EFT and use some smart logic to show FIT/EFT if one these docks not hidden? somehow else?);
* always detect platform for a bank even without combo dirs: need more researching on various images;
* get maximum info about file types and their purpose, to name them more correct and accurate: need more researching;
* detect and validate headers for most of files, based on https://github.com/coreboot/coreboot/blob/main/util/amdfwtool/amdfwtool.h#L384: need more researching;
* much much more testing! I need images, lots of images, the older the better.

This PR is based on:
* Patrick Rudolph's great work: https://github.com/PatrickRudolph/UEFITool/commits?author=PatrickRudolph
* Coreboot's amdfwtool: https://github.com/coreboot/coreboot/tree/main/util/amdfwtool
* some researching and experiments on Genoa/Turin boards.

### DIFF
--- a/UEFITool/uefitool.cpp
+++ b/UEFITool/uefitool.cpp
@@ -603,11 +603,35 @@ void UEFITool::goToAddress()
 {
     goToAddressDialog->ui->hexSpinBox->setFocus();
     goToAddressDialog->ui->hexSpinBox->selectAll();
-    if (goToAddressDialog->exec() != QDialog::Accepted)
-        return;
+    goToAddressDialog->ui->bankComboBox->clear();
+
+    UINT32 address;
+    QModelIndex index;
+    auto indexesAddressDiffs = ffsParser->getIndexesAddressDiffs();
+    if (indexesAddressDiffs.size() < 2) {
+        goToAddressDialog->ui->bankLabel->setVisible(false);
+        goToAddressDialog->ui->bankComboBox->setVisible(false);
+        if (goToAddressDialog->exec() != QDialog::Accepted)
+            return;
     
-    UINT32 address = (UINT32)goToAddressDialog->ui->hexSpinBox->value();
-    QModelIndex index = model->findByBase(address - (UINT32)ffsParser->getAddressDiff());
+        address = (UINT32)goToAddressDialog->ui->hexSpinBox->value();
+        index = model->findByBase(address - (UINT32)ffsParser->getAddressDiff());
+    }
+    else {
+        for (int i = 0; i < indexesAddressDiffs.size(); i++) {
+            index = indexesAddressDiffs.at(i).first;
+            goToAddressDialog->ui->bankComboBox->addItem(model->name(model->parent(index)) + " / " + model->name(index));
+        }
+        goToAddressDialog->ui->bankLabel->setVisible(true);
+        goToAddressDialog->ui->bankComboBox->setVisible(true);
+        if (goToAddressDialog->exec() != QDialog::Accepted)
+            return;
+
+        int selected = goToAddressDialog->ui->bankComboBox->currentIndex();
+        address = (UINT32)goToAddressDialog->ui->hexSpinBox->value();
+        index = model->findByBase(address - indexesAddressDiffs.at(selected).second, indexesAddressDiffs.at(selected).first);
+    }
+
     if (index.isValid()) {
         ui->structureTreeView->scrollTo(index, QAbstractItemView::PositionAtCenter);
         ui->structureTreeView->selectionModel()->select(index, QItemSelectionModel::Select | QItemSelectionModel::Rows | QItemSelectionModel::Clear);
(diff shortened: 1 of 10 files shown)


NikolajSchlej:
Traveling now, will be able to review in 3 weeks, so it will have to sit here for a while.
@vit9696, check this out, meanwhile,

UEFITool/uefitool.cpp
```diff
@@ -48,13 +48,25 @@ version(tr(PROGRAM_VERSION)),
 markingEnabled(true)
 {
     clipboard = QApplication::clipboard();
+    supportedFileTypes = QStringList()
```
vit9696: This looks like an unrelated change. Please move to a separate PR.

UEFITool/uefitool.cpp
```diff
@@ -48,13 +48,25 @@ version(tr(PROGRAM_VERSION)),
...
+        << "*.ffs"
+        << "*.wph"
+        << "*.dec";
 
     // Create UI
     ui->setupUi(this);
     setStyle(new DockProxyStyle(style()));
     setTabPosition(Qt::AllDockWidgetAreas, QTabWidget::North);
     ui->hexViewWidgetContents->layout()->addWidget(&selectedHexView);
-    dockTimer.setSingleShot(true);
```
vit9696, line 57: Why is this removed?

UEFITool/uefitool_main.cpp
```diff
@@ -48,6 +48,7 @@ class UEFIToolApplication : public QApplication
     
     int startup()
     {
+        QApplication::processEvents();
```
vit9696: Why is this added?

common/ffsparser.cpp
```diff
@@ -5431,3 +5441,1178 @@ void FfsParser::outputInfo(void) {
...
+
+    // Full header is part of image?
+    if ((offset + headerSize) > amdImage.size()) {
+        if (!probe)
+            msg(usprintf("%s: ", __FUNCTION__) + typeName + usprintf(" directory table header at %Xh is not within the image", offset), parent);
+        return U_BUFFER_TOO_SMALL;
+    }
+
+    // Fill in table specific details
+    UINT32 checksumStart;
```
vit9696: A lot of code here performs manual parsing. Why do not you use Kaitai?

vit9696 (commented):
Before starting to review this more precisely, I would like

* to move unrelated changes to a separate PR
* to decide whether to use Kaitai for structure parsing

UEFITool/uefitool.cpp
Ykidia: This is old unneeded code, from times when I was experimenting with docks.

common/ffsparser.cpp
Ykidia: I don't know Kaitai. Was reading something but not enough - I have no time for it now.

UEFITool/uefitool_main.cpp
Ykidia: This is needed for Windows (at least) - when opening file in new window, all fonts in most info windows including tree view are reset to some non-mono. After long trying to fix it many ways I found this solution by logic way: when reading settings and setting fonts for whole QApplication we need to allow all widgets to receive signals to really change font.

UEFITool/uefitool.cpp
Ykidia: This affects only 3 places with several lines in 2 files and is cosmetic (no need to write the same long line with the list of extensions several times). This does not deserve a separate PR, but if you want, I can include it in another, cosmetic PR, with other changes.

Ykidia:
PR is updated, mostly to get rid of all unrelated changes. @vit9696, please, check.

NikolajSchlej:
Just returned from a surgery, will take me about a month to recover from it, hope to return to UEFITool development after that. I will review the changes here when capable.

NikolajSchlej:
Tested locally on multiple AMD images, looks really nice and promising, thanks a lot. Merged.

AkechiShiro:
Thanks a lot for the work being done here, really appreciate it,

These changes haven't been released yet into a new release (last release was made June 16 and may not be using the `new_engine` but rather the `old_engine`)

Just would like to test them real quick too, I could probably just build from the HEAD commit on the `new_engine` branch and try it out most likely, is that a recommended approach @NikolajSchlej @Ykidia @vit9696 ?

NikolajSchlej:
I'm planning a release soon, but meanwhile you can build the last commit, as you suggested, and test that way.

Ykidia:
@NikolajSchlej 
You mentioned that you’d need about another month to fully recover, so I figured I’d have time to prepare an improved version of my PR (new commit). The current implementation has a couple of issues I’ve already addressed. 

For example, the relationships between structures aren’t always clear - especially in complex images like for the ASUS ROG STRIX X570, where there are so many interconnections that it becomes hard to understand “who belongs to whom.”
Another issue was the dynamic size of the FET (yes, I’m now consistently using this term, instead of "EFT"), which varies across platform and CPU generations and wasn’t being detected correctly. I’ve already solved both of these problems. 

However, there’s one remaining issue I’m still working on: file attributes. The problem is that file attributes are stored in the table entries, not in the files themselves. But a single file can have multiple parent table entries, which might specify conflicting attributes at the same time.   

Showing attributes in the table entry info makes sense - but hiding them in the file info feels uninformative and ends up being quite inconvenient in practice. So I’m currently figuring out the best way to handle this. 

Would you be okay giving me a bit more time to sort this out? 😊

NikolajSchlej:
Sure. 

As for conflicting attributes, check if that is really the case, and show them both in the table (where they can't conflict) and in the file info (where instead of just "Attributes" for non-conflicting case there will be "Attributes (Bank0)" and "Attributes (Bank1)") or something alike.

PR #448, opened 2025-12-23
Improved AMD-specific structures parsing

Ykidia:
Compared to #443 :
- improved banks detection (but the code is still ugly sometimes);
- directories offsets calculation as for files in physical mode;
- more complete info about items, especially for files;
- enumerating directory table entries for clarity;
- more clear info about item's parents and childs, added bases for each and entry numbers if parent is directory table;
- if a file has several parents then check if some flags and properies values is different, marking such with "*" in info;
- changed "EFT" to "FET", trying to autodetect FET size, more info for FET;
- code refactoring and cleanup.

Tested on ~24 images from different times, but, as usual, much more images needed to explore. In particular, I was never able to find the image with the ISH directory table. If anyone has such images, please, share!

TODOs generally remain the same...


NikolajSchlej:
LGTM

PR #452, opened 2026-01-02
Add method raw() to TreeItem and TreeModel

Ykidia:
raw() = header() + body() + tail()
This is a real helper to make code a little cleaner, simpler and gives more freedom to choose the internal implementation of data storage in TreeItem.

### DIFF
--- a/UEFITool/hexviewdialog.cpp
+++ b/UEFITool/hexviewdialog.cpp
@@ -49,7 +49,7 @@ void HexViewDialog::setItem(const UModelIndex & index, HexViewType type)
     switch (type) {
         case fullHexView:
             dialogTitle = UString("Hex view: ");
-            hexdata = model->raw(index);
+            hexdata = model->entire(index);
             break;
         case bodyHexView:
             dialogTitle = UString("Body hex view: ");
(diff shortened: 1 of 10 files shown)


NikolajSchlej:
I do not like the word 'raw' to mean 'head+body+tail', as 'raw' means 'uncooked' or 'unprepared' or 'unprocessed', and here we have concatenation of up to three UByteArrays as a result of something that means 'nothing had been done with it just yet'.

'whole' maybe?

Ykidia:
> I do not like the word 'raw' to mean 'head+body+tail', as 'raw' means 'uncooked' or 'unprepared' or 'unprocessed', and here we have concatenation of up to three UByteArrays as a result of something that means 'nothing had been done with it just yet'.
> 
> 'whole' maybe?

"raw" because of unstructured nature, in front of structured - header, body, tail, etc. Thought about this a lot and decided that "raw data" was the most appropriate name for me.
But if you dont't like this, we can use something like "entire", "total", "direct". But all these words imply the clarification "data", as it seems to me. But the "data" has a slightly different meaning in model classes...
Honestly, the word "whole" does not highlight the sense of integrated and, simultaneously, unstructured data, but acts as a filler word, in my opinion. But, maybe, "raw" also doesn't describe it completely.
Then we should try "entire" or something else ("total" and "direct" describes norm but sounds bad for me)?

Ykidia:
New idea)! What about "blob" or "plain"?

NikolajSchlej:
'entire' is alright

Ykidia:
Done. Now entire() = header() + body() + tail().

Issue #453, opened 2026-01-03
parseFitEntryAcm: Intel ACM with unknown header version 00050004h found


melvyn2:
I encountered this warning on a Lunar Lake platform SPI dump. Here is the extracted BIOS region: 

[Region_BIOS_BIOS_region.rgn.zip](https://github.com/user-attachments/files/24413229/Region_BIOS_BIOS_region.rgn.zip)

The information panel for this ACM:
```
Fixed: Yes
Base: E7FFE8h
Header address: FFE7FFE8h
Data address: FFE80000h
Offset: 1FFE8h
File GUID: 6520F532-2A27-4195-B331-C0854683E0BA
Type: 01h
Attributes: 04h
Full size: 40018h (262168)
Header size: 18h (24)
Body size: 40000h (262144)
Tail size: 0h (0)
State: F8h
Header checksum: 80h, valid
Data checksum: AAh, valid
```

NikolajSchlej:
Thanks for reporting, will check that out.

NikolajSchlej:
Fixed by https://github.com/LongSoft/UEFITool/commit/8a813b2d6a6263daecc87d5860f1eaecf6a2e6dd

PR #454, opened 2026-01-05
Add basic Brotli Guided Section support

melvyn2:
I copied the way gzip/zlib compression was handled for brotli, including vendoring of code. Some capsule files fail to decompress but those also fail on the reference edk2 tools.


NikolajSchlej:
We delayed adding Brotli because there are zero UEFI images with Brotli sections found in the wild so far. Do you happen to have found some?

melvyn2:
Here's a BIOS region where many of the interesting bits are in a Brotli compressed section: https://github.com/user-attachments/files/24413229/Region_BIOS_BIOS_region.rgn.zip (same dump as #453)

Here is the corresponding capsule file: [P16ALY.zip](https://github.com/user-attachments/files/24436630/P16ALY.zip)
I am still investigating how to transform the strange brotli section data in the capsule into valid brotli data.

melvyn2:
For this capsule specifically, there are 5 blobs of 0x524 bytes inserted into the brotli compressed data at a regular interval. These blobs have a consistent header (`03 00 00 00 05 00 00 00 ...`) and footer but I can't identify them. Once those blobs are removed (but while keeping the brotli section length the same, i.e. taking data from the padding following it), it decompresses normally to the same data.

I have no clue what this is but maybe this can help someone else deal with them.

Edit: These blobs are actually distributed throughout the entire file and so don't have anything to do with this decompressor specifically. They can be removed with a command like the following:
```
perl -0777 -pe 's/\x03\x00\x00\x00\x05\x00\x00\x00.{384}\x01\x00\x01\x00.{384}(\x02\x00\x00\x00.*?\x00\x00\x00\x00\xFF\x00\x00\x00\x00\x00\x00\x00)?//sg' P15ALY.CAP > P15ALY.MOD.CAP
```
The "cleaned" file then works fine with this PR and also works with the Boot Guard parser.

NikolajSchlej:
Thanks for finding that Lunar Lake image, that would be out first Brotli sample found in the wild, and for the PR, of course. Merging.

PR #457, opened 2026-01-12
Fix uniformByte() acts not exactly as when counting bytes if array is empty

Ykidia:
I apologize. The uniformByte() function replacement behaves differently from counting when the array is empty. So I added a wrapper function isUniformByte() with behavior similar to counting.
Previously I have slightly more complex uniformByte() but for https://github.com/LongSoft/UEFITool/pull/451 I tried to simplify and added stupid errors to whole parsing. Now fixed by added simple (I hope) wrapper.
Again, sorry for my negligence.


NikolajSchlej:
No problems, happens to the best of us.

PR #463, opened 2026-02-18
Set of patches from fuzzing campaign 

pagabuc:
This PR contains a set of patches resulting from fuzzing UEFITool. Each commit includes the corresponding crash report (mostly ASAN traces) together with a reference to the crashing input, which are attached.

Based on our testing, these changes do not break any existing functionality. Running the tool against a few hundred firmware images produces identical results with and without this PR. Moreover, the patches improve robustness by enabling successful parsing of additional real-world firmware images. For example, [this HPE firmware](https://downloads.hpe.com/pub/softlib2/software1/sc-windows-uefi-sys/p145084115/v120817/cp030447.exe) sample currently triggers a `std::out_of_range exception`, whereas with this PR it is parsed successfully.

This is joint work with @yeggor, so kudos to him too!

[crashing-inputs.zip](https://github.com/user-attachments/files/25398536/crashing-inputs.zip)

### DIFF
--- a/common/kaitai/kaitaistream.cpp
+++ b/common/kaitai/kaitaistream.cpp
@@ -478,13 +478,24 @@ uint64_t kaitai::kstream::read_bits_int_le(int n) {
 // ========================================================================
 
 std::string kaitai::kstream::read_bytes(std::streamsize len) {
-    std::vector<char> result(len);
-
     // NOTE: streamsize type is signed, negative values are only *supposed* to not be used.
     // https://en.cppreference.com/w/cpp/io/streamsize
     if (len < 0) {
         throw std::runtime_error("read_bytes: requested a negative amount");
     }
+
+    if (len > 0) {
+        // Check that the stream has enough data before allocating
+        std::streampos cur_pos = m_io->tellg();
+        m_io->seekg(0, std::ios::end);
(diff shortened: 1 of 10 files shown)


NikolajSchlej:
Thanks a ton, folks. I'm currently traveling and will be available for reviewing and approving in a week or two.

common/kaitai/kaitaistream.cpp
```diff
@@ -478,14 +478,25 @@ uint64_t kaitai::kstream::read_bits_int_le(int n) {
...
 std::string kaitai::kstream::read_bytes(std::streamsize len) {
-    std::vector<char> result(len);
-
     // NOTE: streamsize type is signed, negative values are only *supposed* to not be used.
     // https://en.cppreference.com/w/cpp/io/streamsize
     if (len < 0) {
         throw std::runtime_error("read_bytes: requested a negative amount");
     }
 
+    if (len > 0) {
```
NikolajSchlej, line 487: Need to indicate that this change in a part of UEFITool-related changes using "//UEFITool: your text here" template. This will ease porting of our changes next time the Kaitai runtime gets an update, and will ensure that we won't lose them.

NikolajSchlej (approved):

PR #464, opened 2026-03-15
test: add Catch2 v3 test harness and Tier 1 unit tests

### LINKED ISSUE #270: No way to continue without a test suite
Long ago, when trees were big and bits were played live, I've developed those parsers in Qt-flavored C++, using lots and lots of different EFI and UEFI images as possible input, and adding a lot of really strange code to make most of them parsing into something resembling a valid tree structure.

Sadly, today in 2022 almost none of those original images are available to re-test that old code, so refactoring it got unnecessary complex, enough to discourage any new people to ever touch anything here. I myself was discouraged enough for many years.

To make this pain a lot easier, we seriously need to:
- write a simple test program that can be executed before and after any code change to confirm that the change did not break anything that already has a test for
- write a ton of tests for this program
- (re)collect some wacky EFI/UEFI images to test the corner cases and document those corner cases in the code (i.e. // This is to parse test image X, Y and Z)
- once all of the above is done, add testing in CI/CD pipeline so robots can do it for us
- then require tests to be written for any new PRs to be merged

FIM43-Redeye:
## Summary

Adds a Catch2 v3 test harness and comprehensive unit tests for all pure-computation functions in `common/`. This is the first step toward addressing #270.

- **Catch2 v3.7.1** vendored as amalgamated single-header into `common/tests/catch2/`
- **Both CMake and Meson** build the same `uefitool_tests` binary
- Test binary compiles as **C++14** (for Catch2 v3); library code remains C++11
- **170 assertions across 33 test cases**, covering:
  - `utility.cpp`: checksums (sum8, checksum8/16/32, fletcher32), uniform byte detection, padding type, pattern matching (makePattern/findPattern), string utilities (fourCC, visibleAsciiOrHex, fixFileName, errorCodeToUString)
  - `ffs.cpp`: uint24/uint32 round-trip conversions, GUID string formatting and parsing, file and section type strings
  - `types.cpp`: itemType, itemSubtype, compressionType, and actionType string conversions (including documenting that `Actions::Erase` falls through to unknown)
  - `descriptor.cpp`: region offset/size calculations, calculateAddress8/16, JEDEC ID lookup

## Design decisions

- One test file per source module (`test_utility.cpp` tests `utility.cpp`, etc.)
- Tests go in `common/tests/` per maintainer direction in #270
- No library source files were modified
- Smoke tests included for each module to catch linkage issues early

## What's next

This covers "Tier 1" -- pure computation functions with no dependencies. Future PRs would add:
- **Tier 2**: TreeItem/TreeModel, GUID database, FFS operations (synthetic data, no firmware images needed)
- **Tier 3**: Integration tests with real firmware images from the community collection

## Test plan

- [x] All 170 assertions pass with CMake (`cmake --build`)
- [x] All 170 assertions pass with Meson (`meson test`)
- [x] No library source files modified
- [x] Existing targets (UEFIExtract, UEFIFind, UEFITool) build without regression

Closes #270 (partially -- test harness + Tier 1)


NikolajSchlej:
Sad, but expected. 

Instead of meaningful testing I'm getting LLM-generated slop that tests trivial functions for being trivially correct. None of this is "testing" or covers anything, and the result of this kind of "testing" is more than useless, as it adds a lot of code, proves absolutely nothing to be correct or wrong, and adds a support burden by making changing the "tested" functions harder.

While I appreciate every new contributor and every new PR from them, this is way below passing my quality gate. Thanks for getting my attention to Catch2, but **LLM slop is not welcome here**. 

I'm closing this PR without merging, repeated attempts at adding code generated by glorified autocompletes to this repository will result only in "llm;dr" remark and a ban.

PR #471, opened 2026-04-17
Revise security policy with incident reporting section

GermanAizek:
Updated the security policy to include incident reporting and response details.


NikolajSchlej:
The tool is not suitable for any security-sensitive applications (almost nothing written in C++ is, TBH).

We consider security issues to be regular bugs, that can be reported using the issue tracker as usual.

There's no need for this PR, declined.

PR #484, opened 2026-07-28
Fix build on OpenBSD/aarch64

brad0:


NikolajSchlej:
Thanks.

Will try to add OpenBSD to the CI/CD pipeline, so breakages like this won't keep happening every time the vendored LZMA SDK gets an update.

brad0:
Well in this case it is specific to aarch64 as opposed to code that is arch agnostic.

Also to note this is already upstream with 26.00 or newer.

NikolajSchlej:
Then there's an LZMA SDK update pending here. I'll accept the PR, but will also update the vendored SDK.

